PR #262, opened 2026-06-12
docs: record agent architecture decisions

onmax:
Records the remaining agent architecture decision notes and framework integration context that still live on the agent-state-provider docs branch.


onmax:
out of scope

PR #357, opened 2026-06-22
refactor(shell): own controlled curl execution

onmax:
Moves controlled Source-backed `curl` execution into the Shell Just Bash provider, so Workspace only passes its Source request executor into Shell instead of owning shell parsing and execution behavior.

This keeps the existing controlled curl feature intact while removing the duplicate Workspace-side shell parser. The Shell provider now advertises network support when Source network grants are present, and Workspace tests alias the Shell workspace source entry so package-local tests exercise the monorepo source instead of stale package output.

Validated with focused Shell and Workspace tests, package typechecks, package builds, and file-scoped lint.


onmax:
@codex review

PR #363, opened 2026-06-22
fix: show agent dev thinking fallback

onmax:
Adds a host-side pending thinking fallback to `vitehub agent dev` so the CLI shows existing trigger `thinkingFallback` metadata, or `Thinking...` by default, until the first visible stream output arrives.

The Vite dev stream now carries trigger invocation metadata on `start`, and the CLI clears the pending fallback before rendering text, tool, usage, delivery, approval, or error events so it never becomes assistant history.


onmax:
@codex review

PR #364, opened 2026-06-23
feat(agent): add named workspace write references

onmax:
## Summary
- keep existing `workspace: "name"` behavior as a read-only named Workspace reference
- add explicit write-capable named references with `workspace: { name, mode: "write" }`
- cover same-backing-store artifact sharing across `runAgent(...)` subagent calls
- add colocated `server/agents/<name>/instructions.md` as the preferred workspace instructions file, materialized as `AGENTS.md` and used as default model instructions only when explicit `driver.instructions` or legacy `instructions` are absent
- keep ordinary Workspace `AGENTS.md` files non-implicit while preserving Source Instructions and Capability instruction composition

## Tests
- `pnpm exec vp test test/workspace.test.ts` from `packages/agent`
- `pnpm typecheck` from `packages/agent`
- `pnpm typecheck` from `docs`
- `git diff --check`

### DIFF
--- a/docs/content/docs/agents/invocations.md
+++ b/docs/content/docs/agents/invocations.md
@@ -90,6 +90,27 @@ export default defineEventHandler(async (event) => {
 
 Agent Trigger Consumers call the trigger surface. They do not own the Capability behavior that registered the trigger.
 
+## Input lifecycle
+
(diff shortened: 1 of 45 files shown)


onmax:
@codex review

PR #365, opened 2026-06-23
feat(agent): add input invocation hook

onmax:
Adds an Agent-level `agent:input` lifecycle hook that runs once after invocation input, invoker, and Capabilities are prepared, but before the selected Agent Driver executes. This gives apps a clean Agent Definition hook for trusted invocation-context guards without attaching fake Capabilities.

Documents the hook on the invocations page and covers the no-Capability regression case with a focused runtime test.


onmax:
@codex review

PR #372, opened 2026-06-24
fix(workspace): prefer executable shell providers

onmax:
Routes Workspace shell inspection through a WorkspaceSession-backed Shell Execution Provider when the input exposes `startSession()`, then through an explicit `executionProvider` hook, and finally through the existing Just Bash virtual filesystem fallback.

This keeps commands such as `rg` as ordinary shell behavior instead of adding command-specific handling, while preserving the current Just Bash path for non-executable Workspaces. Cloudflare shell support has an obvious provider slot now; automatic Cloudflare shell client discovery is still not wired into the Workspace runtime.

Adds source-backed command transcript coverage that replays realistic agent shell commands across lazy sources, controlled `curl`, and native trusted-host sessions. It also covers scoped session propagation so `startSession({ paths })` does not expose unrelated materialized source mounts.


onmax:
@codex review

PR #389, opened 2026-06-26
docs: expand runtime showcase providers

onmax:
Expands the landing-page provider signals and primitive showcase metadata so supported non-default runtimes show up where the examples can honestly render them. Blob now shows Netlify Blobs and Docker/MinIO, KV shows Deno KV and Docker/Node fs-lite, Schedule shows Netlify and Deno, and Workflow shows Docker/OpenWorkflow.

Queue and Sandbox stay Cloudflare/Vercel-only because their package config still only supports those providers. The focused docs tests, docs artifact tests, typecheck, whitespace check, and browser smoke all passed locally before this branch was pushed.


onmax:
@codex review

PR #392, opened 2026-06-26
feat(agent): load colocated agent instructions

onmax:
Loads agent-local `instructions.md` from the same folder as a discovered Agent Definition handler, then injects it as the synthetic workspace `AGENTS.md` source used for default model instructions. This keeps `workspace/instructions.md` as ordinary Workspace content and removes the need for downstream agents such as Quiver Review to add an explicit source or Capability just to mount system instructions.

The generated route wrappers now also preserve already-materialized top-level Agent Definition sources while adding colocated instructions and `sourceRootDir`. This keeps Capability-contributed sources, such as `skills({ source, sourceKey: 'agentBrowserSkill' })`, available to harness workspaces so `skills()` can satisfy its own preflight without duplicate downstream `workspace.sources` entries.

For Quiver, place `instructions.md` next to `server/agents/<agent>/config.ts`. The `workspace/` directory is optional and only needed when the agent has actual Workspace source files. Agent-browser support should continue to use the existing `skills()` Capability with its inline/custom `source`; Quiver should not need a separate `review-instructions` Capability or duplicate skill files under `workspace.sources`.


onmax:
@codex review

PR #395, opened 2026-06-27
fix(database): preserve drizzle registry in server bundles

onmax:
Fixes the database Vite integration so server builds that bundle `@vite-hub/database/drizzle` keep the generated Drizzle Runtime Surface instead of falling through to package import fallbacks. The packed drizzle entry still contains `#vitehub/database/*` package imports; in dependency-bundled builds those can resolve to `dist/runtime/virtual-*.js` before normal `resolveId` sees them, so `hubDb()` now rewrites only the packed drizzle entry to internal virtual ids that the plugin resolves to generated modules.

Adds regression coverage to the database provider-output test by asserting the built SSR server bundle contains named database entries and not the empty `virtual-databases` fallback.


onmax:
@codex review

onmax:
@codex review

PR #396, opened 2026-06-27
feat(agent): let observability include usage telemetry

onmax:
Adds nested capability support so capabilities can declare default dependencies, then uses it to make `observability()` include `usageTelemetry()` unless the user opts out.

This lets observability carry the same Agent Usage Record exposed by usage telemetry, while explicit `usageTelemetry(...)` configuration still wins when present. The docs now call out the default, the opt-out path, and that `observability.usage` is an alias for the same usage record rather than a second record to persist.


onmax:
@codex review

onmax:
@codex review

PR #435, opened 2026-06-28
feat(agent)!: trust channel route metadata after admission

onmax:
Adds `route.input.trust` for authenticated Channel routes so `webChat({ route })` and custom Channel definitions can copy already-authenticated `meta`, `user`, and `session` request fields into `chat.message` input without a custom mapper. The trust list is only honored after `admission.authenticate`, and `run`, `invoker`, and `invokerProfileId` stay server-owned.

Also moves generated `createChannel*RouteHandler` factories behind `@vite-hub/agent/server/internal` and the matching Vite facade so `defineChannel()` and Channel helpers remain the public source of truth for invocation route admission. The docs now separate Channel delivery/admission from Chat state and history.


onmax:
@codex review

PR #440, opened 2026-06-28
feat(agent): remove entry capability

onmax:
Removes the legacy `entry()` Capability export and docs so app-owned ingress uses Channel triggers instead.

Updates trigger documentation and regression coverage for custom Channel triggers, the missing `entry` export, hook-owned transient reactions, and GitHub pull-request comment reply finish effects. `inputCommands()` remains the command admission and rewriting surface while the GitHub Channel owns webhook delivery, normalized PR context, and write-back effects.


onmax:
@codex review

PR #442, opened 2026-06-28
fix(agent): materialize scoped harness workspace files

onmax:
Fixes harness workspace preparation for Access-scoped runs by reusing the final selected Workspace Scope after Capability Workspace Contributions, instead of recomputing all source mount roots. This keeps generated capability files such as pull-request-context/context.md materialized without widening the harness session to the whole lazy source mount.

Also skips write-back mkdir calls for directories already present in the initial harness tree, so unchanged source-backed mount roots do not trip read-only write policy during session close.


onmax:
@codex review

PR #448, opened 2026-06-29
fix(agent): upstream bitacora harness patches

onmax:
Moves the downstream Bitácora pnpm patches into ViteHub source. Harness-backed Agent Drivers can now set `driver.permissionMode` (`allow-reads`, `allow-edits`, `allow-all`) while raw harness `permissions` stay unsupported; the default remains `allow-all`.

Also initializes the GitHub Workspace Store baseline during load and makes `diff()` load before reading that baseline, so fresh GitHub-backed harness workspace sessions do not see the whole remote tree as added.


onmax:
@codex review

PR #452, opened 2026-06-29
fix(agent): pass thread history to chat webhooks

onmax:
Passes persisted Chat SDK thread messages into `chat.message` webhook runs when chat history is configured with `source: "thread"`. This keeps adapters like Telegram from storing history without the model receiving it on follow-up turns.

Adds a webhook regression that sends two messages through the test Telegram adapter with `threadHistory` enabled and asserts the second run receives the prior user turn, bot reply, and current user turn.


onmax:
@codex review

onmax:
@codex review

PR #455, opened 2026-06-30
feat(workspace): add standalone nitro config export

onmax:
Adds `@vite-hub/workspace/nitro` as the standalone Nitro config entrypoint for `createWorkspaceNitroConfig()` while keeping it a narrow re-export of the existing Workspace Vite helper. Generated Workspace Nitro plugins now export a plain Nitro plugin function, so Nitro 3 apps do not need the old `nitropack/runtime` shim.


onmax:
@codex review

PR #459, opened 2026-07-04
feat(agent)!: rename chat trigger history option

onmax:
Renames the chat trigger input window from `messages.history` to `messages.triggerHistory` and fails legacy history config with a migration error. The runtime now derives a thread-backed trigger window from explicit `threadHistory.maxMessages` when no override is set, while `triggerHistory: "none"` remains the stateless path and explicit `triggerHistory` can diverge from the thread cache size.

Updates Agent package types/exports, webhook and route handling, regression coverage, docs, and `.agents` language so thread cache/backfill stays separate from the model-facing Chat History Window.


onmax:
@codex review

onmax:
@codex review

PR #461, opened 2026-07-05
fix(agent): handle unsupported attachment-only chat messages

onmax:
## Summary
- add a text fallback when chat webhook messages contain only unsupported attachments
- keep audio attachment parts behind the existing includeAudioAttachments option
- cover image-only Telegram-style updates in provider tests

## Test
- pnpm test -- providers.test.ts


onmax:
@codex review

onmax:
@codex review

PR #464, opened 2026-07-06
fix(agent): serialize sqlite state connect

onmax:
## Summary

- make the SQLite Agent State Provider share concurrent `connect()` calls
- skip reconnect work when the adapter is already connected
- add a regression test for concurrent connect attempts using the existing injected driver surface

## Root Cause

Discord Gateway startup can forward `READY` and `GUILD_CREATE` into the same Agent Definition at the same time. Both Chat instances can call the same SQLite state adapter's `connect()` concurrently. The adapter set `connected = true` before migration/cleanup, but did not guard duplicate migrations, so local libSQL could hit `SQLITE_BUSY` or a later cleanup could observe a half-connected state.

## Validation

- `corepack pnpm exec vp test run test/sqlite-state-provider.test.ts` from `packages/agent`
- `corepack pnpm exec vp run -t @vite-hub/agent#build`
- `git diff --check`

### DIFF
--- a/packages/agent/src/state/sqlite.ts
+++ b/packages/agent/src/state/sqlite.ts
@@ -150,6 +150,10 @@ export class ViteHubSqliteAgentStateAdapter implements StateAdapter {
   }
 
   async connect(): Promise<void> {
+    if (this.connectPromise) {
+      await this.connectPromise
+      return
+    }
     if (this.connected) return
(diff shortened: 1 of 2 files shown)


onmax:
@codex review

onmax:
@codex review

PR #468, opened 2026-07-06
fix(agent): format dev loop stream errors

onmax:
Formats Agent Dev Loop and invocation-stream failures with structured details instead of letting non-Error values collapse to `[object Object]`.

The root cause was direct string interpolation of stream error payloads and thrown values. This keeps Error message/stack behavior, emits JSON for structured object payloads, and falls back without exposing `[object Object]` for unserializable values.

Validated with focused `@vite-hub/agent` build, CLI/invocation-stream tests, typecheck, and `git diff --check`.

### DIFF
--- a/packages/agent/src/agent-error.ts
+++ b/packages/agent/src/agent-error.ts
@@ -21,6 +21,40 @@ function isError(error: object): boolean {
   }
 }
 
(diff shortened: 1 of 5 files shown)


onmax:
@codex review

PR #469, opened 2026-07-06
fix(agent): honor webhook runtimes

onmax:
## What changed

Quiver carried a dist-level patch for `@vite-hub/agent` that swapped the channel webhook path from `runAgentInline` to `runAgent`. This moves that fix upstream.

Webhook-triggered non-chat Agent Invocations now go through `runAgent`, so an Agent Definition with `runtime: workflow(...)` starts a Workflow Run. Agents without a workflow runtime still fall back to inline through `runAgent`.

The shared HTTP response helper now preserves safe Workflow Run fields (`id`, `provider`, `status`, optional `metadata`/`result`) and omits `payload` from JSON responses.

## Call sites

Changed: non-chat `createChannelWebhookRouteHandler` Agent Trigger Consumer. Bypassing `runAgent` there ignored the Agent Definition runtime.

Left inline: chat webhook handling and DevTools/live chat streaming paths. Those paths post or stream platform responses inside the chat adapter flow, and the existing non-streaming chat webhook workflow test still asserts inline behavior.

## Validation

- `pnpm --filter @vite-hub/agent test`

### DIFF
--- a/packages/agent/src/http-response.ts
+++ b/packages/agent/src/http-response.ts
@@ -17,7 +17,7 @@ function isAsyncIterable(value: unknown): value is AsyncIterable<unknown> {
 const workflowProviders = new Set(["cloudflare", "openworkflow", "vercel"])
 const workflowRunStatuses = new Set(["completed", "failed", "queued", "running", "unknown"])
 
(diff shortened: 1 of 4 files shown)


onmax:
@codex review

onmax:
@codex review

onmax:
@codex review

PR #472, opened 2026-07-06
fix: upstream quiver dependency patches

onmax:
Upstreams the Quiver-carried ViteHub dependency patches now that #471 has merged: Agent Channel Delivery Effect contexts expose the originating `Request`, and Cloudflare R2 Blob Stores use the runtime binding when present or fall back to lazy `files-sdk/r2` HTTP access when no binding is available.

The Blob resolver now preserves R2 HTTP options and masks env-sourced runtime credentials so generated Provider Output can rehydrate them without committing secrets.

Refs #471

PR #473, opened 2026-07-06
docs: document Cloudflare R2 HTTP fallback

onmax:
Documents the Cloudflare R2 Blob Store behavior added in #472: runtime bindings are preferred, and HTTP credentials can back fallback access when no binding exists.

Adds the real config/env names and optional R2 HTTP dependency install command to the Blob docs.

Refs #472


onmax:
@codex review

PR #478, opened 2026-07-06
fix(agent): pass rest capability cli argv as input

onmax:
Moves the Quiver-carried Capability CLI rest-command behavior into Agent Package source. Commands marked `rest: true` now receive leftover argv as `input.argv` instead of parsing those args as Capability CLI flags, while preserving the command path and explicit JSON mode.

Adds focused runtime coverage for a rest command receiving `--filter` as command input.


onmax:
@codex review

PR #479, opened 2026-07-06
fix(agent): allow blob puts from workspace files

onmax:
Adds `workspacePath` as an alternate `blob_edit` put body so agents can upload a Workspace file directly to Blob storage while still requiring exactly one of `body` or `workspacePath`.

Blob capability tools now resolve the Blob primitive only when a tool executes, using the runtime primitive when present and falling back to an installed `@vite-hub/blob` runtime export when no primitive was injected.


onmax:
@codex review

PR #480, opened 2026-07-06
fix(agent): quiet invocation stream abort cleanup

onmax:
Upstreams the Quiver-carried invocation stream abort cleanup into Agent Package source. Agent Invocation Stream cancellation now treats `AbortError` from abort cleanup as non-fatal, and the Vite Agent Invocation Stream Endpoint cancels response bodies once on a later tick while ignoring abort-shaped body failures.

Adds focused tests for cancellation abort cleanup and Vite response `AbortError` handling.


onmax:
@codex review

PR #471, opened 2026-07-06
fix: avoid static vercel blob peer reachability

onmax:
Removes ViteHub-owned static reachability to the Vercel Blob optional peer so apps that do not use Vercel Blob do not pull `files-sdk/vercel-blob` into build analysis.

The legacy Blob bundled driver now delegates to the existing lazy Vercel driver, and the Workspace Vercel Blob store keeps its `files-sdk/vercel-blob` import behind a non-static optional-peer path. Focused build-output tests cover both surfaces.

PR #481, opened 2026-07-06
fix(workspace): remove files-sdk from vercel blob store

onmax:
Removes the Workspace Vercel Blob Store's `files-sdk/vercel-blob` delegation and calls the selected `@vercel/blob` runtime peer directly through the existing lazy hosted-store boundary.

Vercel Blob workspace build output now materializes only `@vercel/blob`, and focused tests assert the built internal store has no `files-sdk` edge or static `@vercel/blob` import.

Refs #471


onmax:
@codex review

PR #482, opened 2026-07-06
fix(agent): map text chat attachments

onmax:
Ports the Bitacora-carried chat text attachment patch into Agent Package source. Chat SDK file attachments with text-like MIME types or txt/md/json/yaml/log/csv filenames now become text parts before chat access and agent execution, with a 1 MiB cap.

The Agent Package also exposes the Discord Gateway route handler through `@vite-hub/agent/server`, and generated Discord Gateway routes import that public handler so apps can rely on the ViteHub-owned route instead of carrying their own Nitro route.

Adds provider coverage for text attachment mapping, oversized text attachment rejection before chat access, GET-only Discord Gateway admission, and the gateway-started webhook path for Discord `.txt` attachments.

### DIFF
--- a/packages/agent/src/server.ts
+++ b/packages/agent/src/server.ts
@@ -1,3 +1,7 @@
+export {
+  createDiscordGatewayRouteHandler,
+} from "./server/routes.ts"
+
 export type {
   AgentChannelChatRouteAdmissionContext,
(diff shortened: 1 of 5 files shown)


onmax:
@codex review

PR #483, opened 2026-07-06
fix(workspace): fall back to github store without hosted loader

onmax:
Upstreams the Bitacora-carried `@vite-hub/workspace` generated-dist patch at the Workspace Store boundary. GitHub-backed Workspace Stores now instantiate the internal GitHub Store directly when no hosted store loader is installed, while Cloudflare Artifacts and Vercel Blob still require runtime loader wiring.

Adds a focused lifecycle regression covering `createWorkspaceStore()` with a GitHub provider and no hosted loader.


onmax:
@codex review

PR #487, opened 2026-07-07
fix(agent): normalize git tool inputs

onmax:
## Summary
- accept string, cmd, and argv-style git tool inputs in addition to command
- normalize bare git subcommands like diff --stat to git diff --stat
- keep rejecting shell composition such as && and |

## Tests
- pnpm exec vp test run test/git-capability.test.ts test/invocation-stream.test.ts --reporter=dot
- pnpm --dir packages/agent typecheck


onmax:
@codex review

PR #491, opened 2026-07-07
fix(agent): ignore generated harness tool writeback

onmax:
Prevents the generated `harness-tool.mjs` helper from being written back into a Workspace after harness-backed Agent runs. The helper is runtime scaffolding, so it now joins the existing harness instruction files in the writeback ignore list.

Validated with the focused Agent workspace test and typecheck.


onmax:
@codex review

PR #493, opened 2026-07-07
fix(workspace): prefer source keys for github mounts

onmax:
Updates Workspace GitHub source normalization so named source bindings use the binding key as their default mount while explicit mounts still win.

Numeric or opaque source keys still fall back to the GitHub repo slug, and the existing default-ref behavior is covered with a non-main default branch test.


onmax:
@codex review

PR #519, opened 2026-07-09
fix(agent)!: expose usage telemetry as capability

onmax:
Keeps usage telemetry as an official `usageTelemetry()` capability and exposes end-of-invocation usage through the existing finish extension getter: `event.extensions.get("usage-telemetry")` and delivery `context.extensions.get("usage-telemetry")`.

The extension returns flat primitive JSON (`UsageTelemetryRecord`) with token, speed, cost, response, model, and run fields when available. ViteHub does not format comments, messages, or UI; apps own formatting. The temporary public `event.usage` and `context.usage` fields are removed.

Research note: `.agents/docs/context-usage-record.md`


onmax (commented):
make sure to use document writer skill and update the docs before merging

PR #524, opened 2026-07-09
fix(agent): require explicit input command rewrites

onmax:
Removes the handlerless args passthrough added in #521 so accepted Input Commands without explicit handler text fall through to the existing command-text removal path. That keeps app-owned handlers responsible for model-facing prompt text instead of sending slash command args as a vague prompt.

Updates the focused Agent Package tests for handlerless `/review ...` and `/summary`, and adjusts the Input Commands docs to say model-facing text must be returned explicitly.

Refs #521


onmax:
@codex review

PR #525, opened 2026-07-09
fix(agent): preserve prefixed data stream events

onmax:
Preserves `data-*` Agent Invocation Stream events when normalizing agent output, so chat-title data events and similar UI data chunks survive alongside text and finish events.

This keeps the public stream/message event type aligned with the data-prefixed events ViteHub already emits and adds a focused agent-output regression for the downstream Bitacora escape.


onmax:
@codex review

onmax:
@codex review

PR #533, opened 2026-07-10
fix(agent): trust chat route timeouts after admission

onmax:
Adds `timeout` to authenticated Channel route input trust so app-owned chat routes can forward caller-selected Agent Invocation timeouts after admission. Only positive, finite numeric values are copied; invalid, unlisted, and unauthenticated timeout values never reach the Agent Driver.

This replaces the downstream Quiver patch that forwarded timeout values from every generated route, while preserving the Channel Delivery Admission boundary introduced in #435. The public docs and runtime/type coverage now describe and protect that contract.

### DIFF
--- a/docs/content/docs/agents/channels.md
+++ b/docs/content/docs/agents/channels.md
@@ -111,7 +111,7 @@ export default defineAgent({
           },
         },
         input: {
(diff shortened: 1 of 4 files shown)


onmax:
@codex review

PR #534, opened 2026-07-10
fix(agent): run workflow-backed evals inline

onmax:
Agent Evals use `createAgentTestRunner()`, which previously delegated to `runAgent()` and therefore dispatched workflow-backed Agent Definitions through `@vite-hub/workflow`. That made local evals start workflow execution and forced downstream callers to reconstruct private Agent options without `runtime`.

The test runner now uses the existing inline Agent Invocation path. A focused regression proves eval/test execution neither creates nor runs a workflow handle, while ordinary `runAgent()` behavior remains workflow-backed.

### DIFF
--- a/packages/agent/src/test.ts
+++ b/packages/agent/src/test.ts
@@ -1,6 +1,6 @@
 import {
   defineAgent,
-  runAgent,
+  runAgentInline,
   withAgentDefaults,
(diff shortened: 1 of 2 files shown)


onmax:
@codex review

PR #538, opened 2026-07-10
fix(agent): expose channel context and align workspace scopes

onmax:
Introduces `channel` as the canonical Agent Invocation Context value for message-shaped Channels, carrying the current message, metadata, run, session, and user. Safe registered invocation values are exposed directly to capability callbacks, dynamic model metadata callbacks, and Workspace Source resolvers, while legacy and internal aliases remain available only through the invocation context store. The public types and docs cover application-owned channel metadata and resolver augmentation.

Also aligns Workspace Source scoping so unscoped Sources are available in every selected Workspace Scope, source-declared scope names can be selected without duplicate empty Access definitions, and explicitly scoped Sources remain restricted to matching names. Focused runtime and type coverage protects the callback paths, alias filtering, resolver context, Access grants, and scoped overlays.

This PR is temporarily based on `agent/actor-email` because both changes touch the same Chat input and type-test hunks. It should be retargeted to `main` after the actor-email PR lands.

### DIFF
--- a/docs/content/docs/agents/channels.md
+++ b/docs/content/docs/agents/channels.md
@@ -90,7 +90,11 @@ export default defineEventHandler(async (event) => {
 })
 ```
 
-The `run` fields are first-class Agent Run metadata, not Chat context. They help DevTools, traces, and finish hooks explain where the invocation came from.
+The `run` fields are first-class Agent Run metadata, not legacy Chat context. They help DevTools, traces, and finish hooks explain where the invocation came from.
+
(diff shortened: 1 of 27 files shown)


onmax:
@codex review

onmax:
@codex review

onmax:
@codex review

onmax:
@codex review

onmax:
@codex review

onmax:
@codex review

PR #540, opened 2026-07-10
feat(agent): resolve capability CLIs per invocation

onmax:
Allows Capability CLI contributions to resolve from the current Agent Invocation while keeping the Capability statically attached to its Agent Definition. The OpenAPI Capability now uses that public resolver instead of the private `resolveCli` convention, and `openapi({ cli })` can return CLI options, `false`, or `undefined` without loading the OpenAPI spec for omitted invocations.

This lets multi-channel Agents expose a Portal-only CLI without making Channels own Capabilities or treating Channel metadata as authorization. Trusted request, Agent Actor, and `access()` checks remain the authority boundaries; runtime, generated OpenAPI CLI, Agent Dev Loop, and public-type coverage protect the new behavior.


onmax:
@codex review

PR #541, opened 2026-07-10
fix(agent): expose channel context and align workspace scopes

onmax:
Adds a canonical typed `channel` Agent Invocation Context Value for message-shaped Channels and exposes registered invocation values directly to Capability callbacks, dynamic model metadata callbacks, instruction composition, and Workspace Source resolvers. Legacy and internal aliases remain available through `context.get(...)` without leaking into the direct callback object.

Aligns Workspace Source scope selection with that context path: unscoped Sources are available in every selected scope, scoped Sources match by name, and named Source scopes no longer require duplicate empty `access.workspace.scopes` entries. DevTools inputs, documentation, runtime and type contracts, and focused regression coverage are updated together.

### DIFF
--- a/docs/content/docs/agents/channels.md
+++ b/docs/content/docs/agents/channels.md
@@ -90,7 +90,11 @@ export default defineEventHandler(async (event) => {
 })
 ```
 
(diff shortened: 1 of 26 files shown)


onmax:
@codex review

PR #542, opened 2026-07-10
fix(agent): allow parallel local bridge sessions

onmax:
Default local harness sessions advertised fixed port `4000`, while the harness lease registry only coordinates sessions that share the same provider object. Invocation-scoped sandbox callbacks can return distinct providers, so concurrent bridge starts collided with `EADDRINUSE` and one Claude Code run exited before becoming ready.

This lets the operating system assign each default local bridge port atomically by advertising `ports: [0]` and omitting the provider fixed-port pool unless the caller explicitly supplies `ports`. It adds a focused regression for parallel provider instances, preserves explicit fixed pools, and captures the primary-source evidence in the accompanying research note.

### DIFF
--- a/.agents/research/local-sandbox-parallel-bridges.md
+++ b/.agents/research/local-sandbox-parallel-bridges.md
@@ -1,0 +1,41 @@
+# Parallel local harness bridge ports
+
+## Decision
+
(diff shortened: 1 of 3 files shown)


onmax:
@codex review

PR #543, opened 2026-07-10
feat(agent): ship Valibot schema default

onmax:
Ships Valibot through `@vite-hub/agent/valibot` as the optional zero-install default for Agent tool and Capability CLI schemas. The Standard Schema boundary remains validator-agnostic, so applications can still use Zod, ArkType, or another compatible validator directly.

Adds the package export, runtime dependency, build entry, and a public-import contract test. The custom Capability documentation now uses the built-in Valibot path for both direct tool inputs and Capability CLI schemas.


onmax:
@codex review

PR #549, opened 2026-07-11
docs: make host support inspectable

onmax:
Turns the framework and host documentation into a source-backed support contract. It adds a matrix that separates app-facing runtime helpers, local providers, generated provider output, provisioning, and production proof across Local Vite, Cloudflare, Vercel, Netlify, Deno, Nitro/UnJS, and Node/self-hosted. The host guides now spell out provider-owned boundaries and current proof levels instead of implying universal portability.

It also brings reference and development documentation back in sync with package contracts: accurate @vite-hub/vite defaults, discovery and file conventions, public import paths, provisioning credentials, and generated output. A package README and regression tests preserve those boundaries, while duplicate development copies of the Agent eval and devtools pages are removed so each topic has one canonical owner.


onmax:
@codex review

PR #555, opened 2026-07-11
fix(agent): preserve native chat title delivery

onmax:
Moves `chatTitle()` stream decoration onto a reader-backed `ReadableStream` merger. The previous async-generator wrapper changed `result.stream` from a `ReadableStream` into a plain async iterable, which broke result getters that call `pipeThrough()`. It could also delay cancellation behind a pending provider read; cancellation now reaches the source reader immediately and preserves the disconnect reason.

The Agent Dev Loop now renders title effects as `[delivery] title …`, alongside reactions, while retaining the raw payload for malformed or custom title effects. Provider-native Discord title mutation is proposed separately in [vercel/chat#693](https://github.com/vercel/chat/pull/693), so applications using lazy official adapter factories do not need app-local REST wrappers.

Validated with the full `@vite-hub/agent` suite (42 files, 1,023 tests), type checking, focused Oxlint, package build/publint, and a patch-first Bitacora reproduction of the original `pipeThrough()` failure.

### DIFF
--- a/packages/agent/src/capabilities/chat-title.ts
+++ b/packages/agent/src/capabilities/chat-title.ts
@@ -370,10 +370,10 @@ function withChatTitleReadableStreamParallel<T>(
   title: Promise<string | undefined>,
   renderTitle: (title: string) => T,
 ): AsyncIterable<T> & ReadableStream<T> {
-  const reader = result.getReader()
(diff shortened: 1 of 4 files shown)


onmax:
@codex review

onmax:
@codex review

PR #556, opened 2026-07-11
feat(agent): allow trusted-host command execution

onmax:
Adds an explicit `workspaceShell({ commands: "trusted-host" })` mode for Linux-hosted agents that need to discover and run executables such as `gh`, `git`, interpreters, and generated scripts without an application-maintained allowlist. The capability refuses this mode unless the resolved Workspace runtime is also `trusted-host`; existing executable arrays and sandbox behavior remain unchanged.

Execution keeps the structured `command` plus `args` contract, Workspace-relative cwd validation, environment restrictions, timeouts, session cleanup, and write-mode commits. The underlying trusted-host runtime still rejects production unless the Workspace explicitly sets `allowProduction: true`. Documentation and DevTools state the boundary directly: this is the host service account's filesystem, network, credential, and process authority, not isolation.

Validated with the full `@vite-hub/agent` suite (42 files, 1,023 tests), package type checking, build/publint, focused Oxlint, strict review, `git diff --check`, and the existing real-session Workspace production-guard tests.

### DIFF
--- a/docs/content/docs/capabilities/workspace-shell.md
+++ b/docs/content/docs/capabilities/workspace-shell.md
@@ -1,6 +1,6 @@
 ---
 title: Workspace shell
-description: Expose shell-shaped Workspace inspection, mutation, and allowlisted Workspace command tools.
+description: Expose Workspace inspection, mutation, and explicit executable tools.
 navigation.title: Workspace shell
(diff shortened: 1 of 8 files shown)


onmax:
@codex review

PR #557, opened 2026-07-11
fix(agent): scope generated chat state per agent

onmax:
Scopes framework-provided Chat SDK state by agent and Channel across cache, locks, queues, subscriptions, and Channel-local history. Generated LibSQL and Cloudflare state was previously shared by every discovered agent, so identical provider message IDs could let one agent suppress another through the shared dedupe key.

Transcript lists are scoped per agent rather than per Channel so an explicit cross-Channel `identity` still provides one shared transcript. Explicit `messages.state` remains untouched. Host state resolvers also remain unwrapped because they already receive `chat.stateKeyPrefix` and own their namespace.

Direct framework state now uses `chat:${agentName}:${origin}:` for Channel-local records and `chat:${agentName}:transcripts:user:${userKey}` for transcripts. This changes the generated-state key layout, so existing transient dedupe, queue, lock, subscription, history, and transcript records are not reused after the upgrade.

### DIFF
--- a/docs/content/docs/capabilities/chat-title.md
+++ b/docs/content/docs/capabilities/chat-title.md
@@ -43,6 +43,8 @@ export default defineAgent({
 `chatTitle()` runs in the output phase.
 It wraps compatible async streams or UI message streams so the title can arrive alongside the response, and it provides `{ title }` in the finish extension.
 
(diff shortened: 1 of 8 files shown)


onmax:
@codex review

onmax:
@codex review

PR #560, opened 2026-07-11
feat(agent): capture invocation data by default

onmax:
Makes invocation observability core data instead of an opt-in capability. Every Agent Invocation now receives a per-invocation metadata-only Trace Event Log and trace context when the host did not supply them; caller-provided trace objects and sinks remain authoritative. Agent Finish events expose normalized `resultKind` and `usage`, and terminal trace events carry the same completion facts without persisting content.

Agent test results and eval observations now expose a finalized `TraceRunView`. `observability()` and `usageTelemetry()` remain compatible but are deprecated and consume the core finish data first, giving applications a migration path without another capability layer.

Completion-correct tracing wraps bare `Response` and async-iterable outputs, so object identity is intentionally not preserved. Payloads, status, headers, streaming, cancellation, and error propagation remain covered and preserved. The full Agent suite passes (42 files, 1,025 tests), along with docs tests/typechecks/build, Agent typecheck/build/publint, focused Oxlint, strict review, and `git diff --check`.

### DIFF
--- a/docs/content/docs/agents/evals.md
+++ b/docs/content/docs/agents/evals.md
@@ -95,20 +95,20 @@ Use `--watch` while editing prompts or scenarios. Use `--threshold` and `--outpu
 ## Score useful behavior
 
 Good evals score source-grounded answers, refusal behavior, expected tool use, no source leakage, and regressions in usage or latency when Agent usage exists.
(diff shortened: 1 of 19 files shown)


onmax:
@codex review

onmax:
@codex review

onmax:
@codex review

onmax:
@codex review

PR #561, opened 2026-07-11
refactor(agent)!: replace agent defaults with runtime identity

onmax:
Replaces generated-host definition cloning with an explicit runtime identity boundary. Discovered Nitro, Netlify, Deno, Discord Gateway, Vite DevTools, and invocation-stream hosts now pass `AgentHostIdentity` through `context.agentIdentity`, so the same source Agent Definition can run under multiple host identities without mutation or hidden defaults.

Workflow identity resolves explicit binding → definition name → host identity. Implicit Workspace identity resolves explicit reference or string → definition name → identity workspace → identity name → the built-in default. Identity is preserved through Workflow payloads, while source-root and colocated-instruction preparation remain unchanged.

This is a breaking cleanup of the competing identity model: `withAgentDefaults`, `AgentHandlerOptions`, and hidden `__vitehubWorkspaceAgentDefaults` metadata are removed. Custom hosts invoking discovered Agent Definitions must pass `agentIdentity` at runtime.

PR #564, opened 2026-07-11
refactor(agent)!: remove agent defaults

onmax:
## Summary

- remove the public `withAgentDefaults()` helper, `AgentHandlerOptions`, and hidden `__vitehubWorkspaceAgentDefaults` metadata
- make explicit `defineAgent({ name })` and runtime `agentIdentity` the only identity paths
- delete redundant DevTools/default plumbing while keeping workspace registration payloads isolated from returned definitions

## Context

This breaking draft is stacked on #561, which first moves generated hosts, workflows, tests, and DevTools to discovered runtime identity. Keeping the old cloning helper after that migration would preserve two competing identity models.

Custom hosts now pass `agentIdentity` when invoking a discovered Agent. `registerWorkspaceAgent()` returns the original definition and registers a source-root-enriched workspace payload; a regression test invokes that returned definition with runtime identity and reads its colocated workspace file.

## Validation

- `corepack pnpm --filter @vite-hub/agent test` — 42 files, 1022 tests
- Agent build and publint
- Oxlint on the changed runtime and tests — no errors
- `git diff --check`

The diff is net deletion: 189 additions / 323 deletions. The only non-implementation references to `withAgentDefaults` are negative assertions proving it is absent from public and generated output.

### DIFF
--- a/docs/content/docs/concepts/definitions-and-discovery.md
+++ b/docs/content/docs/concepts/definitions-and-discovery.md
@@ -15,7 +15,7 @@ Location-derived identity makes generated registries, Provider Output, DevTools,
 
 This matters for agents because an Agent File Name or agent folder name becomes the discovered Agent identity. `defineAgent({ name })` is not the discovery identity override.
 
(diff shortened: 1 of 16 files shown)


onmax:
@codex review

PR #554, opened 2026-07-11
feat(schedule): support IANA time zones

onmax:
Adds an optional IANA `timeZone` to Runtime Schedule records, create/update inputs, persistence, and runner matching. Schedules without a time zone continue to use UTC, while zoned schedules follow local civil time across daylight-saving transitions.

Exposes the same field through the Agent Schedule Capability and updates the public package and documentation examples. Static Schedule Definitions and Provider Wake output remain UTC-only.

PR #565, opened 2026-07-12
feat(agent): expose one cronjob tool

onmax:
Replaces the separate `schedule_read` and `schedule_edit` tools with one `cronjob` tool. Read mode exposes `targets`, `list`, and `get`; write mode adds `create`, `edit`, `pause`, `resume`, `run`, and `delete`, including immediate execution through the existing Runtime Schedule client.

Read operations remain automatically allowed while mutating operations retain the configured policy and default approval requirement. Target allowlists and explicit self-target permission continue to apply to every operation.

## Coordination

This PR and #554 can merge independently, but both touch the Schedule Capability implementation, tests, and docs. The later merge will need a focused rebase.

### DIFF
--- a/docs/content/docs/capabilities/schedule.md
+++ b/docs/content/docs/capabilities/schedule.md
@@ -1,6 +1,6 @@
 ---
 title: Schedule
-description: Declare Agent Schedules or expose Runtime Schedule tools to an Agent.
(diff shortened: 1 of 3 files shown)


onmax:
@codex review

PR #566, opened 2026-07-12
docs(workspace): clarify trusted-host service account boundary

onmax:
Documents the trusted-host service-account boundary and locks the existing environment inheritance contract with a real Workspace Session regression test. Commands run as the ViteHub service account and inherit `HOME`, XDG and GitHub CLI configuration, the SSH agent socket, and other process secrets, so tools such as `git`, `gh`, and `ssh` can reuse that account's setup.

Clarifies that read mode controls Workspace tools and writeback rather than host side effects, and that Workspace rules and Workspace Scope do not sandbox arbitrary host processes. Also corrects the Workspace runtime reference to include trusted-host execution and its explicit `allowProduction` opt-in.

Validated with the complete `@vite-hub/workspace` suite (25 files, 374 tests), package type checking and build/publint, the documentation production build, focused Oxlint, strict review, and `git diff --check`.


onmax:
@codex review

PR #567, opened 2026-07-12
feat(schedule): add runtime inputs and targets

onmax:
Adds opaque, generic `input` to Runtime Schedule records, create/update calls, and handler context. Built-in memory and KV stores snapshot input deeply; providing input on update replaces the whole value, while omitting it preserves the stored snapshot.

Adds `defineScheduleTarget({ handler })` for cronless Runtime Schedule targets. These definitions are runtime-eligible by construction and remain discoverable through the runtime registry and target list, while static Nitro, Cloudflare, Vercel, Netlify, and Deno provider output ignores them. `defineSchedule()` remains cron-required.

This PR is stacked on #554 so the combined Runtime Schedule record contract keeps both `timeZone` and generic input.


onmax:
@codex review

onmax:
@codex review

PR #572, opened 2026-07-12
feat(agent): improve finish hooks and chat errors

onmax:
Allows `agent:finish` hooks to return reply, reaction, and status delivery effects, which are delivered after capability finish effects. Agent definitions can add post-run channel behavior through the same delivery contract as capabilities instead of wiring channel-specific callbacks.

Chat webhooks now preserve the message from a canonical `RateLimitRejectedError`, so applications can give users useful retry guidance while unrelated failures still use the generic fallback. The default, configured, and spoofed-error paths are covered by regression tests.

### DIFF
--- a/docs/content/docs/agents/invocations.md
+++ b/docs/content/docs/agents/invocations.md
@@ -173,6 +173,19 @@ export default defineAgent({
 })
 ```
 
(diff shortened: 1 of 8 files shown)


onmax:
@codex review

PR #576, opened 2026-07-13
fix(schedule): clean unused provider output

onmax:
Avoids creating empty provider output when no static Schedule definitions are emitted, and removes stale ViteHub-owned Netlify and Vercel functions plus generated schedule artifacts when static schedules are disabled.

Vercel cleanup only removes schedule-owned cron paths and deletes the default config/output when no sibling provider content remains. Unrelated provider files, functions, config fields, routes, and crons are preserved. Cloudflare array-owned cleanup also stops creating an absent output root.


onmax (commented):
make sure this netlify folder only appears when netlify provider is used

onmax:
@codex review

onmax:
@codex review

PR #577, opened 2026-07-13
feat(agent): add trusted-host boxes

onmax:
Adds `@vite-hub/box` as the provider-neutral definition and resolved runtime boundary, with one concrete `trustedHost()` runtime. Agent Definitions can inline a mutable checkout, optional managed Home, and named boot requirements; `codexDriver()` contributes Codex automatically and uses the host profile, authentication, global skills, and CLI configuration while harness cache state stays disposable.

Connects Box resolution directly to harness Agents and rejects conflicting Workspace or driver sandbox wiring. Focused tests cover definition resolution and trusted-host execution, the docs distinguish Box, Workspace, and Sandbox ownership, and the real Babysitter `runAgent()` path completes against the existing ChatGPT, GitHub CLI, Git, and pnpm setup without dirtying the target checkout.

### DIFF
--- a/docs/content/docs/agents/agent-definitions.md
+++ b/docs/content/docs/agents/agent-definitions.md
@@ -5,7 +5,7 @@ navigation.order: 21
 icon: i-lucide-file-user
 ---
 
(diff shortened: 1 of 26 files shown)


onmax:
@codex review

PR #579, opened 2026-07-13
fix(schedule): log process runtime errors

onmax:
Logs Process Runtime failures to stderr before forwarding the same normalized `Error` to Nitro's capture hook. A failing capture hook remains isolated, so systemd and other long-running hosts retain actionable failures in their ordinary service logs.

This closes the visibility gap found in Babysitter, where failed Schedule runs were durable but absent from journald.

Verdict: proceed

What is good: keeps the fix scoped to generated Nitro Process Runtime error visibility; logs the same normalized Error that is forwarded to Nitro capture; leaves capture-hook failures isolated.

What is missing: no broader runtime inspection surface, but that is outside this pull request intent.

Corrections applied: none.


onmax:
@codex review

PR #582, opened 2026-07-13
feat(agent): accept channel factories in definitions

onmax:
Allows synchronous zero-argument Channel factories to be registered directly in `defineAgent({ channels })`. ViteHub resolves each factory once per Agent Definition and keeps normalized Channel objects at runtime, so aliases, generated routes, delivery, and discovery continue using the existing contract.

Explicit Channel definitions, `stream()`, and configured helpers such as `stream(options)` remain supported. The public input types, Channels guide, type coverage, runtime isolation, invalid-result validation, and generated-route equivalence are included in the same change.

Verdict: proceed

Good: The PR keeps the external Agent Definition API obvious by accepting synchronous zero-argument Channel factories while normalizing them once into existing Channel definitions.

Missing: No blocking direction gap found; exact-head readiness still depends on implementation review, checks, and review evidence.

Corrections applied: None.


onmax:
@codex review

PR #583, opened 2026-07-13
fix(deps): remove obsolete h3 patch

onmax:
Updates H3 from `2.0.1-rc.20` to `2.0.1-rc.25` and removes the pnpm patch that let the deprecated H3 v2 `getRequestHeaders()` helper accept H3 v1-style plain header objects.

The patch was added for docs prerendering, but the docs now resolve H3 v1 directly and generate successfully without it. Keeping the fallback would hide an invalid H3 v1/v2 event boundary, which upstream H3 intentionally declined to support in [h3js/h3#1439](https://github.com/h3js/h3/pull/1439).

Verdict: proceed

Good: the PR removes the obsolete H3 v2 compatibility patch and advances the UnJS H3 catalog pin while preserving the existing docs-only H3 v1 compatibility catalog.

Missing: no code changes are required for the validated direction; exact-head implementation review and external review evidence still remain separate readiness gates.

Corrections applied: none.


onmax:
@codex review

PR #587, opened 2026-07-14
fix(blob): infer filesystem base from environment

onmax:
Allows filesystem Blob stores to resolve their base from `BLOB_FS_BASE` when `blob.base` is omitted, while keeping explicit config precedence and the `.data/blob` fallback. This makes deployment-owned mount paths configurable without app-local Vite config or package patches.

Documents the config-time precedence and covers implicit resolution, explicit filesystem stores, and explicit override behavior.


onmax:
@codex review

PR #590, opened 2026-07-14
fix(vite): make optional integrations opt-in

onmax:
Stops `vitehub()` from registering KV, Sandbox, and Schedule unless the application requests them, aligning their activation with the existing opt-in Queue integration. Passing `true` uses each package's inferred defaults, explicit options configure the integration, and omission or `false` leaves it inactive.

Previously these integrations were enabled implicitly, so applications had to carry negative `false` configuration for primitives they did not use. The preset tests and package/reference documentation now capture the positive activation contract; package tests, typecheck, build, and documentation tests pass.


onmax:
@codex review

PR #595, opened 2026-07-14
fix(agent): infer observability finish extensions

onmax:
Registers the existing observability finish extension in ViteHubAgentFinishExtensions, so known Agent Finish Extension reads infer AgentObservabilityFinishExtension and its usage field without explicit generics. Unknown capability ids and fields keep the existing unknown fallback.

This preserves event.invocation.usage as the preferred normalized usage API and leaves observability runtime behavior and deprecation unchanged. Type-level regression coverage captures whole-extension, field, and invalid lookup inference.

### DIFF
--- a/packages/agent/src/capabilities/observability.ts
+++ b/packages/agent/src/capabilities/observability.ts
@@ -67,6 +67,12 @@ export interface AgentObservabilityFinishExtension {
   usage?: AgentUsageRecord
 }
 
+declare global {
+  interface ViteHubAgentFinishExtensions {
+    observability: AgentObservabilityFinishExtension
+  }
+}
+
 interface AgentObservabilityCapabilityMetadata<
   TRuntimeConfig extends AgentRuntimeConfig = AgentRuntimeConfig,
   CALL_OPTIONS = unknown,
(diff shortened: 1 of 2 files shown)


onmax:
**Head:** `ff7c83c5c0c651a6d9e6de72802db2f821e9b08a`
**Verdict:** `revise` (correction applied)

**Direction:** Register `observability` as `AgentObservabilityFinishExtension` in the existing Agent Finish Extension registry, update every affected type reader, and preserve optional presence plus unknown fallbacks.

**Evidence:** The Capability already returns and exports this shape; `AgentInvocationExtensions` derives known-key inference from `ViteHubAgentFinishExtensions`; the stale eval observation assertion now expects the registered type. The branch is rebased onto current `main` (`819d2687`); Agent typecheck passes and focused runtime/eval tests pass (268 tests).

**ViteHub precedent:** `usage-telemetry` registers its built-in finish extension through the same adjacent global declaration in `packages/agent/src/capabilities/usage-telemetry.ts`.

**Strongest challenge:** This is global registry inference, not per-Agent installed-capability inference. That broader known-key behavior is the established registry invariant and remains safe because reads are optional.

**Next action:** Force-push under the verified prior-head lease, wait for checks, then request one Codex review for this exact rebased head.

onmax:
@codex review

onmax:
@codex review

PR #598, opened 2026-07-14
refactor(agent)!: remove stream channel

onmax:
Removes the public `stream()` Channel factory and `"stream"` kind, and makes `webChat()` the route-enabled default for AI SDK UI-message-stream HTTP chat. This keeps Channels destination-named while runtime and HTTP streaming APIs remain unchanged.

Resolves `messages.stream` from the active adapter-backed Channel so progressive delivery is channel-scoped. Tests, docs, and the Deno and Netlify fixtures now use `webChat()` directly; other channel-local message settings retain their existing single-message-Channel limitation.

### DIFF
--- a/docs/content/docs/agents/channels.md
+++ b/docs/content/docs/agents/channels.md
@@ -13,21 +13,20 @@ Channel Kind helpers are imported from `@vite-hub/agent/channels`, not the root 
 
 ```ts [server/agents/support.ts]
 import { defineAgent } from '@vite-hub/agent'
-import { github, stream, webChat } from '@vite-hub/agent/channels'
+import { github, webChat } from '@vite-hub/agent/channels'
 
 export default defineAgent({
   channels: {
     github: github({ pullRequest: true }),
-    portal: stream,
-    web: webChat(),
+    portal: webChat,
   },
   run: () => 'ok',
 })
 ```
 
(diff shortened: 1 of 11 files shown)


onmax:
**Head:** `32537f8a3c8fe34a552e00088356c1703ddb23d3`

**Verdict:** `proceed`

**Direction:** Remove the generic Stream Channel, make Web Chat route-enabled by default, and resolve adapter streaming from the active destination Channel; Web Chat HTTP streaming remains independent of adapter delivery.

**Evidence:** The one-commit diff deletes the public `stream()` factory/options and `"stream"` kind, defaults `webChat()` routes on with `route: false` as the opt-out, overlays the active Channel's `messages.stream` during webhook dispatch, and keeps every other multi-Channel message override rejected. Public docs, type tests, runtime tests, provider tests, Deno/Netlify fixtures, and repository CI cover the contract. The Vercel preview failure is an explicit daily deployment quota while the repository Vercel verification check passes.

**ViteHub precedent:** Channel helpers are destination/provider-named (`webChat`, `github`, `slack`, etc.); Agent Invocation and HTTP streaming already have separate runtime primitives. Generated route admission already belongs to Channel definitions.

**Strongest challenge:** Default-enabling a generated route changes existing `webChat()` behavior, but the PR explicitly requires that breaking final-design contract, provides `route: false`, and avoids retaining a compatibility alias that would preserve two Channel pattern families.

**Next action:** Run focused verification and strict base-branch diff review, fix only actionable findings, then request one exact-head Codex review.

onmax:
@codex review

PR #604, opened 2026-07-15
fix(agent): allow opted-in capability tools by default

onmax:
Treats attaching or configuring a write-enabled Capability as the developer's explicit opt-in by letting omitted tool policy resolve to `allow` across KV, Blob, DB, Git, Memory, Repository Host, and Schedule. Developers can still set `require-approval` or `deny` when the product needs an additional runtime gate.

Preserves the Capability-owned security boundaries around write modes, store and schema selection, Git command restrictions, memory scopes, and Schedule target allowlists, then aligns the tests, public types, and documentation so official Capabilities no longer require `policy: 'allow'` boilerplate.

The Agent and Runtime suites, docs build, package typechecks, lint, workspace builds, and contract tests pass. The full root verification still reaches five unrelated existing `@vite-hub/box` failures caused by macOS path normalization and BSD tar behavior.

### DIFF
--- a/docs/content/docs/capabilities/blob.md
+++ b/docs/content/docs/capabilities/blob.md
@@ -43,7 +43,7 @@ export default defineAgent({
 
 ViteHub selects the configured Blob store and exposes a small Storage Capability Tool Surface.
 Read mode supports one object read, metadata read, or prefix list operation per tool call.
-Write mode adds put/delete operations with approval required by default.
+Write mode adds put/delete operations and allows them by default.
 Put operations accept either `body` or `workspacePath`, not both.
 Delete operations return `{ pathname, deleted: true }`.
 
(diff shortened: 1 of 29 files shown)


onmax:
**Head:** `7156c14de95df4d80350f7a5497d3a9d0d239fbe`

**Verdict:** `proceed`

**Direction:** Attaching or configuring a write-enabled Capability opts the Agent into its bounded write tools, so omitted tool policy resolves to `allow`; use `require-approval` or `deny` for an additional Runtime gate.

**Evidence:** The Agent tool runtime already executes tools directly when policy is absent. This diff removes Capability-specific `require-approval` fallbacks without removing explicit write modes, per-store write opt-ins, Git command restrictions, SQL/schema checks, memory scopes, store selection, Schedule target allowlists, or input validation. Implementation, tests, public docs, and Runtime documentation align; focused Runtime and Agent tests pass, Agent typechecking passes, and regression assertions cover explicit mixed-store Memory policy plus Schedule read/write dispatch.

**ViteHub precedent:** Read operations and other model-facing tools represent the normal allowed path by omitting policy. Keeping authority selection in Capability configuration and policy as an optional later Runtime gate follows ViteHub's existing package boundary and makes the obvious API match runtime behavior.

**Strongest challenge:** Autonomous-write defaults are security-sensitive and deliberately break the prior approval-by-default convention. The change remains coherent because the developer must explicitly enable each write surface and the Capability's operation-level boundaries still apply before policy.

**Next action:** Wait for checks on the pushed exact head, request one exact-head Codex review, and address any terminal findings before merge.

onmax:
@codex review

PR #610, opened 2026-07-15
fix(docs): center and repair search

onmax:
Centers the docs search modal by removing the legacy horizontal transform that now duplicates Tailwind's translate utility. Upgrades Docus to 5.12.3 and enables its FTS5 search path so valid documentation queries return real pages instead of `No matching data`.

Verified at a 1600×900 desktop viewport: the 640px dialog is centered at `x=480`, searching for `workspace` returns documentation results, and selecting `Workspace shell` navigates to `/docs/capabilities/workspace-shell/`. The docs test suite, typecheck, production build, and frozen-lockfile install pass.

PR #611, opened 2026-07-15
feat(docs): refine navigation card grids

onmax:
## Summary

Documentation card grids now use Nuxt UI's `pageGrid` and `pageCard` app-config slots instead of stale class selectors. The shared treatment keeps Docus and Nuxt Content's existing MDC components, while presenting a compact two-column desktop index, a single-column mobile stack, subtle dividers, and matched dark-mode states.

## Testing

- `pnpm --filter vitehub-docs typecheck`
- `pnpm --filter vitehub-docs test` — 31 tests passed
- `pnpm --filter vitehub-docs build`
- Browser QA on the docs overview, card navigation, mobile navigation, dark mode, and reference-table overflow

The local runtime proof used the Docus 5.12 update from #610 plus a temporary resolution pin for Nuxt Content's current H3 mismatch. Neither dependency workaround is included here, so this PR stays focused on the UI layer.

### DIFF
--- a/docs/app/app.config.ts
+++ b/docs/app/app.config.ts
@@ -85,6 +85,22 @@ export default defineAppConfig({
     pageBody: {
       base: "mt-0 px-4 pb-24 sm:px-8 lg:px-8 xl:px-12",
     },
+    pageGrid: {
+      base: "grid-cols-1 gap-px border border-default bg-[var(--ui-border)] sm:grid-cols-2 lg:grid-cols-2",
+    },
+    pageCard: {
+      defaultVariants: {
+        variant: "ghost",
+      },
+      slots: {
+        root: "group rounded-none bg-default",
+        container: "gap-0 p-4 sm:p-5",
+        leading: "mb-3 size-8 items-center justify-center border border-default bg-muted/30",
+        leadingIcon: "size-4 text-muted transition-colors group-hover:text-highlighted",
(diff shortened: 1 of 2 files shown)


onmax:
**Head:** `466bdc9cbee87daa82c52bbcdf3c3f8d2105e1de`

**Verdict:** `proceed`

**Direction:** Configure all docs navigation-card grids through Nuxt UI’s `pageGrid` and `pageCard` app-config entries, and remove the stale `.u-page-*` descendant CSS.

**Evidence:** The exact two-file diff moves component presentation into `docs/app/app.config.ts`, where this docs app already centralizes Nuxt UI slots and variants, and deletes nine lines of generated-class CSS. Visible docs/deployment checks pass; there are no reviews or unresolved threads.

**ViteHub precedent:** Existing header, page, navigation, search, and prose components are customized through the same `ui` app-config boundary. `main.css` remains for bespoke prose structures that lack a stable component slot.

**Strongest challenge:** App-config defaults apply to every page grid/card in the docs app, broader than the old `.docs-content` selectors. That reach is intentional because the PR body specifies a shared treatment, and the app config itself is docs-scoped.

**Next action:** Keep the two-file scope, verify typecheck/tests/build and exact responsive/theme behavior, run strict review against `main`, then obtain exact-head Codex review.

onmax:
@codex review

PR #622, opened 2026-07-15
feat(blob): add signed R2 requests

onmax:
Adds `blob.sign(pathname, options)` as the provider-neutral contract for short-lived private object access. Signed requests return the method, URL, and required headers; Cloudflare R2 supports signed `GET` and `PUT`, binds `Content-Type`, and maps `createOnly` to `If-None-Match: *`, while unsupported drivers fail explicitly.

R2 signing uses its S3-compatible HTTP credentials even when normal Blob reads and writes use a Workers binding. This lets downstream applications delete their AWS SDK signing adapter while keeping direct uploads conditional and source downloads short-lived, with matching runtime, public-type, package, and documentation coverage.

### DIFF
--- a/docs/content/docs/server-primitives/blob.md
+++ b/docs/content/docs/server-primitives/blob.md
@@ -179,6 +179,7 @@ Objects from the served store include a URL. With `serve.publicBaseUrl`, the URL
 | `blob.head(pathname)` | Reads object metadata. |
 | `blob.list(options?)` | Lists objects with optional `prefix`, `limit`, `cursor`, and folded folders. |
 | `blob.del(pathnames)` | Deletes one or more objects. |
+| `blob.sign(pathname, options)` | Signs a short-lived `GET` or `PUT` request for one object. |
 | `blob.serve(event, pathname)` | Serves an object stream through an H3 event. |
 | `blob.store(name)` | Selects a named Blob Store. |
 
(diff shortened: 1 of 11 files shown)


onmax:
**Head:** `26d85dfe851edf0e03e5d2ecd435ddc60bc00450`
**Verdict:** proceed

**Direction:** Add `blob.sign(pathname, options)` as the provider-neutral signed-request primitive, with Cloudflare R2 GET/PUT support, explicit unsupported-driver failures, public types, documentation, and focused tests.

**Evidence:** The PR title/body and exact-head diff consistently implement that contract across Blob storage, runtime forwarding, the R2 driver, package metadata, docs, and tests. The branch is rebased onto current `main`; focused signing tests and Blob typecheck pass.

**ViteHub precedent:** Blob already owns provider-neutral storage operations and delegates provider behavior through `BlobDriverAdapter`; an optional driver signing capability exposed as a required storage method follows that boundary and keeps AWS/R2 details behind ViteHub language.

**Strongest challenge:** AWS SDK v3 defaults can add an empty-payload CRC32 checksum to presigned PUT requests created without a body, causing ordinary non-empty R2 uploads to fail. Exact head configures `requestChecksumCalculation: "WHEN_REQUIRED"` and asserts it at the signing boundary.

**Next action:** Await checks and the exact-head Codex review, address any actionable result, then merge when all completion conditions hold.

onmax:
@codex review

PR #628, opened 2026-07-16
fix(provider-output): preserve Nitro Vercel functions

onmax:
Preserves Nitro's Vercel server output when Blob, Database, Queue, and Workflow provider builds run alongside `nitro:main`. Composed builds now emit product-owned sidecar functions without replacing Nitro's function or routes, while standalone package builds retain their existing `__server.func` contract.

Centralizes esbuild filtering for unsupported trailing-slash Vite aliases and Node export-condition selection, so every provider bundle uses the same safe resolution boundary.

### DIFF
--- a/packages/blob/src/internal/vite-build.ts
+++ b/packages/blob/src/internal/vite-build.ts
@@ -45,6 +45,7 @@ interface GenerateProviderOutputsOptions {
   clientOutDir: string
   providerOutput?: ComposedProviderOutput
   rootDir: string
+  serverFunctionName?: string
 }
 
 interface GeneratedBlobArtifacts {
@@ -331,7 +332,11 @@ function createCloudflareOutput(blob: BlobModuleOptions | ResolvedBlobModuleOpti
   }
 }
 
-function createVercelOutput(artifacts: GeneratedBlobArtifacts, providerOutput: ComposedProviderOutput | undefined): VercelProviderDeploymentOutput {
+function createVercelOutput(
+  artifacts: GeneratedBlobArtifacts,
+  providerOutput: ComposedProviderOutput | undefined,
+  serverFunctionName?: string,
+): VercelProviderDeploymentOutput {
   const databaseRuntime = getProviderRuntimeModule(providerOutput, "database", "vercel")
 
   return {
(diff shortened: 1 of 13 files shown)


onmax:
**Head:** `1872e54774b87c13c38d2e4d93ec483032300c7a`

**Verdict:** proceed

**Direction:** Preserve Nitro's Vercel Build Output API function and routes during composed ViteHub provider builds by giving Blob, Database, Queue, and Workflow product-owned sidecar functions only when Nitro owns Vercel hosting; retain the existing `__server.func` contract otherwise. Keep provider bundling resolution safe for Vite aliases and Node conditional exports.

**Evidence:** The current diff selects product-specific function names only when `nitro:main` is present and Vercel hosting is declared or detected through the Nitro preset, ViteHub hosting, or Vercel runtime environment. An explicit Nitro preset is authoritative, so ambient Vercel variables cannot override a non-Vercel preset. This avoids output-file hook-order races. `writeProviderDeploymentOutputs` then uses an explicit `serverFunctionName` and empty `config`, preserving Nitro's existing function/routes. Missing plugin lists and non-Vercel Nitro builds retain standalone behavior. Node esbuild bundles set only the `node` custom condition, leaving `import` and `require` contextual.

**ViteHub precedent:** Shared host-independent output mechanics live in `@vite-hub/internal`, while Blob, Database, Queue, and Workflow contribute product artifacts through the same provider deployment-output primitive. Existing provider normalization already uses ViteHub hosting and provider runtime environment to select deployment behavior.

**Strongest challenge:** Neither Nitro plugin identity nor generated-file existence alone proves stable Vercel ownership, and ambient CI environment must not contradict an explicit Nitro preset. The repair makes explicit configuration authoritative, avoids hook timing, and tests non-Vercel Nitro plus contextual CommonJS export resolution.

**Next action:** Exact-head CI passes and Codex found no major issues; complete final timeline/thread preflight and merge.

onmax:
@codex review

onmax:
@codex review

onmax:
@codex review

onmax:
@codex review

PR #629, opened 2026-07-16
fix(vite-hub): publish Env config types

onmax:
Publishes the Env Vite module augmentation from the `vite-hub` root declaration, so applications that import `vitehub()` from `vite-hub` can use the top-level Vite `env` config without a downstream pnpm patch.

The declaration bundler dropped the runtime-only `@vite-hub/env/vite` import. The package build now preserves that declaration dependency, with a built-package consumer fixture covering `vite-hub` together with `vite-hub/env`.

### DIFF
--- a/packages/vite-hub/fixtures/published-types/consumer.ts
+++ b/packages/vite-hub/fixtures/published-types/consumer.ts
@@ -1,0 +1,12 @@
+import { defineConfig } from "vite"
+import { vitehub } from "vite-hub"
+import { env } from "vite-hub/env"
+
+export default defineConfig({
+  env: {
+    server: {
+      GH_TOKEN: env(),
+    },
+  },
+  plugins: [vitehub()],
+})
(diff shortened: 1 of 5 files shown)


onmax:
**Head:** `530a82f3c3ba5869ab8973e2cebc4f4b89695e2f`
**Verdict:** proceed

**Direction:** Preserve the Env capability's Vite module augmentation in the published `vite-hub` root declaration and verify it from the built package boundary.

**Evidence:** `packages/vite-hub/src/index.ts` imports `hubEnv` from `@vite-hub/env/vite`, but declaration bundling removes that runtime-only import. The built declaration now begins with that augmentation import, and the consumer fixture typechecks `defineConfig({ env: ... })` through `vite-hub` and `vite-hub/env`.

**ViteHub precedent:** `packages/schedule/vite.config.ts` uses a declaration-bundle `generateBundle` hook to preserve declarations that the bundler cannot infer, with a built-package consumer test.

**Strongest challenge:** Editing `index.d.ts` by filename in a build hook is coupled to bundler output. No existing less-coupled declaration-entry primitive is present; the Schedule package establishes this as the current local mechanism.

**Next action:** Await exact-head CI, obtain exact-head Codex review, address any terminal findings, and merge when all evidence is positive.

onmax:
@codex review

PR #633, opened 2026-07-16
fix(agent): keep Evalite subpath-only

onmax:
Removes Evalite from the root `@vite-hub/agent` optional peers, restoring the subpath-only integration contract. The framework distribution had reintroduced the peer, which kept Evalite’s AI SDK 6 graph reachable during downstream upgrades even when applications never import `@vite-hub/agent/eval`.

Evalite remains a development dependency for Agent Package tests, and the existing eval documentation continues to require applications using the eval subpath to install `evalite` and `vitest` explicitly. The package contract test now protects that boundary.


onmax:
**Head:** `0f1e5ce878c018e8b0ba7ad9bbde20d9a0b37b2d`
**Verdict:** proceed

**Direction:** Keep Evalite as an Agent Package development dependency for tests and the eval implementation, but remove it from the root `@vite-hub/agent` peer contract so only consumers of `@vite-hub/agent/eval` opt into installing it.

**Evidence:** The PR changes only `packages/agent/package.json` and its package-contract assertions. Evalite remains in `devDependencies`; `src/eval.ts` and the eval runner own the integration; the eval docs and CLI docs explicitly install `evalite` and `vitest`; the branch is one commit ahead of and fully based on current `main`.

**ViteHub precedent:** Optional integrations are exposed through explicit Agent Package subpaths, and the import-path reference lists `@vite-hub/agent/eval` as the Agent Eval authoring boundary. The root package should not make a subpath-only tool part of every consumer dependency graph.

**Strongest challenge:** The Vite plugin and CLI can invoke Evalite without a direct application import of the eval subpath. However, the documented eval workflow already requires explicit installation, while retaining Evalite as a root optional peer makes its incompatible AI SDK graph visible to all root-package consumers.

**Next action:** Run the focused package-contract test and Agent Package verification, review the exact diff against `main`, then obtain an exact-head Codex review before merging.

onmax:
@codex review

PR #648, opened 2026-07-16
refactor(agent): use canonical Workspace Source metadata

onmax:
Deletes the Agent package's 323-line Workspace Source metadata mirror and moves every caller to the canonical `@vite-hub/workspace` contract, including Access grants, capability contributions, harness preparation, and DevTools metadata. The Workspace package now exposes the narrow `source-metadata` entry point needed by Agent without widening its main surface.

It also removes the earlier probe deferral, which only postponed an invalid absolute File Source. Global Skill directories now require a directory-capable `github()` or `custom()` Source, or a root-confined `glob()` Source, while root-relative File Sources remain single-file; real Workspace reads cover directory materialization, nested references, root-relative single-file mounting, direct and nested absolute-path rejection, and cyclic binding rejection.

PR #632, opened 2026-07-16
chore: refresh workspace dependencies

onmax:
Refreshes the workspace catalogs and lockfile to current compatible releases, with targeted major upgrades for the dependency owners behind the deprecated graph. Narrow root resolutions move Drizzle Kit, glob, UUID, and `parse-url` redundant `@types/parse-path` stub off deprecated versions without adding compatibility wrappers.

This removes eight of the ten deprecated locked versions. `node-domexception@1.0.0` and `prebuild-install@7.1.3` remain because the latest `files-sdk` Google-auth path and maintained native addons still depend on them, with no non-deprecated compatible upstream release.

The refreshed graph also exposed an existing `unknown` type on the Agent Vite plugin devtools extension. The extension now derives its concrete type from `chatDevTools`, preserving contextual typing without changing runtime behavior.

Verified with a frozen install, full workspace typecheck, and all 1,230 Agent tests.

PR #655, opened 2026-07-16
fix(database): restore Drizzle generation compatibility

onmax:
Drizzle Kit 1.0.0-rc.4 imports `drizzle-orm/_relations`, which stable Drizzle ORM 0.45.2 does not export. That breaks `drizzle-kit generate` for ViteHub’s generated D1 HTTP configs.

This restores the compatible 0.31.10 line, removes the global rc override, and adds an executable D1 HTTP generation regression. The Database package suite passes all 77 tests, typecheck reports no errors, and the package build and publint checks are clean.

The current hosted result is green for docs, consumer, package preview, and all three provider-verification jobs. Root CI passes 1,241 tests and only fails four inherited global Skill Workspace cases on `/opt/skills/{code-review,review}`; ready and fully green #648 fixes those cases. The standalone Vercel deployment is separately blocked by the account build-rate quota. There is no Drizzle-specific hosted failure.

### DIFF
--- a/packages/database/test/cli.test.ts
+++ b/packages/database/test/cli.test.ts
@@ -1,10 +1,20 @@
+import { execFile } from "node:child_process"
+import { mkdtemp, readFile, readdir, rm, writeFile } from "node:fs/promises"
 import { delimiter, join } from "node:path"
+import { promisify } from "node:util"
 
-import { describe, expect, it, vi } from "vitest"
+import { afterEach, describe, expect, it, vi } from "vitest"
 
 import { createDbCliContributor } from "../src/cli.ts"
 
 import type { ResolvedDBViteConfig } from "../src/types.ts"
+
+const execFileAsync = promisify(execFile)
+const tempDirs: string[] = []
+
+afterEach(async () => {
+  await Promise.all(tempDirs.splice(0).map(dir => rm(dir, { force: true, recursive: true })))
+})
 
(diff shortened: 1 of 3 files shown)


onmax:
**Head:** `e876d7e62f1177f1c3f9fa86c6aa18e7e432432e`
**Verdict:** proceed
**Direction:** Restore the compatible stable Drizzle Kit 0.31 line for the Database package and retain an executable regression proving the generated ViteHub D1 HTTP shape can generate a migration.
**Evidence:** PR #632 broad dependency refresh changed only Drizzle Kit from `^0.31.10` to `1.0.0-rc.4` while leaving stable `drizzle-orm ^0.45.2`; rc.4 imports `drizzle-orm/_relations`, which that stable ORM does not export. This PR removes the global rc override and its test executes `drizzle-kit generate` with the same `d1-http` contract produced by `packages/database/src/internal/generated.ts`.
**ViteHub precedent:** Before #632, the database catalog used `drizzle-kit ^0.31.10`; ViteHub Database docs and generated config deliberately support D1 HTTP.
**Strongest challenge:** Pinning back could conceal an intended Drizzle 1.0 migration, but no matching ORM prerelease/API migration or PR intent exists, and keeping incompatible release lines breaks the existing public workflow.
**Next action:** Verify the focused package suite and lockfile, review the exact diff against `main`, fix actionable findings, then obtain exact-head Codex review and merge when checks and threads are clean.

onmax:
@codex review

PR #666, opened 2026-07-17
refactor(rate-limit): use source-local handles

onmax:
Replaces Rate Limit Definition files and string-based runtime lookup with source-local handles:

```ts
const uploads = defineRateLimit("image-upload", {
  limit: 5,
  window: "1m",
})

await uploads.consume(key)
```

The Vite integration now collects static top-level declarations through the compiler AST, reports duplicate IDs with source locations, and generates provider output without a registry or file convention. This intentionally removes the transitional definition helpers, updates the Agent adapter and public examples, cleans up legacy generated registry files, and requires a deployment-unique Cloudflare namespace so counters cannot leak across Workers or environments.

PR #667, opened 2026-07-17
feat: add request-aware upload primitives

onmax:
Adds an explicit @vite-hub/blob/content-type detector for common image and PDF signatures, including provider-output aliases so the subpath works in Cloudflare and Vercel bundles. Blob writes remain storage operations; applications opt into byte classification before upload.

Makes managed Rate Limit handles request-aware: .consume() uses the host integration client request key, while .consume(key) remains the explicit user or tenant override. Nitro installs the key through global middleware so the async context covers the route lifecycle, trusts the Cloudflare direct-ingress address only on a Cloudflare host with an H3 fallback, and fails instead of sharing an unknown production bucket.

Stacked on #666 because this extends the source-local Rate Limit handle introduced there.


onmax:
**Head:** `ad614d243a65400429f6a9ce4ffd32ddb6104b8f`

**Verdict:** proceed with correction

**Direction:** Keep the two explicit, upload-oriented primitives in the current title/body: opt-in Blob byte classification and request-aware managed Rate Limit consumption. Complete the public package boundary and avoid broader upload workflow APIs.

**Evidence:** The exact-head diff adds `@vite-hub/blob/content-type` with focused signature tests and provider aliases, makes `RateLimitHandle.consume()` derive the request key while preserving explicit keys, and now forwards the detector through `vite-hub/blob/content-type`. The focused framework contract passes 7/7 with no type errors. PR #666 remains open as the stated source-local handle base.

**ViteHub precedent:** Feature APIs are owned by `@vite-hub/*` packages and forwarded through matching `vite-hub/*` subpaths; host-specific request plumbing stays behind the Vite/Nitro integration.

**Strongest challenge:** “upload primitives” combines Blob classification with generic Rate Limit request identity, but the PR body explicitly scopes and connects both behaviors, while each implementation remains a reusable server primitive.

**Next action:** Await exact-head CI, obtain the required exact-head Codex review, address any actionable findings, and merge when all gates are satisfied.

onmax:
@codex review

onmax:
@codex review

onmax:
@codex review

onmax:
@codex review

PR #676, opened 2026-07-18
refactor(schedule): run process wakes with Effect

onmax:
The Process Schedule Wake Driver now owns polling, deduplication, bounded delivery, and shutdown through one scoped Effect runtime. Its ManagedRuntime boundary uses `runPromiseExit` and explicit Exit/Cause triage, so typed failures, defects, and interruption reasons retain their original identity and no Effect `FiberFailure` reaches callers.

The driver factory and Promise contracts remain unchanged, but integrated #687 intentionally tightens the public ScheduleError constructor and JSON contract. `new ScheduleError(message, options?)` now requires a stable existing `code`; direct callers and subclasses must migrate to `new ScheduleError(message, { code, details?, cause?, httpStatus?, requestId?, retryable? })`. This break is deliberate: keeping the code optional would preserve the unstable message-only contract, while inventing a fallback code would make the typed failure model dishonest. `toJSON()` now exposes only code, safe details, message, requestId, and retryable; causes, stacks, httpStatus, arbitrary invalid values, handler input, provider bodies, and URLs remain internal.

This is stacked on #670 at `4898061`, which owns the Agent Effect dependency, exact beta.99 version policy, and generated-project `msgpackr-extract: false` rule. The integration preserves the externally amended queue-drain, aggregate engine-floor, and cadence fixes. It also includes merged #684 retained-plain OpenWorkflow backend lifecycle and #685 retained-plain worker shutdown, where the Promise/Map generation owner, cached stop result, and synchronous host callback bridge remain plain because Effect Scope would not delete that machinery.

At merged exact head bc9e811254851524b8917db905faeacf7b2e9847, the complete Schedule surface passes build, publint, typechecking, and its full test surface, including TestClock ownership, raw defect and caller-cancellation identity, error redaction and serialization, installed-package declaration coverage, and Effect/FiberFailure public leak guards. GitHub aggregate CI, consumer, docs, package preview, provider verification, Continuous Releases, Workers, and the Vercel deployment are green.

DEPENDENCY STATUS
  base: #670 at 4898061ba242a046e15c52249b72b488073037f4
  integrated: #684 retained-plain OpenWorkflow backend lifecycle
  integrated: #685 retained-plain OpenWorkflow worker shutdown
  integrated: #687 stable ScheduleError contract
  downstream: #681 must rebase onto this final head

EFFECT ADOPTION STATUS
  seam: Process Schedule Wake Driver and ScheduleError public contract
[...]

PR #678, opened 2026-07-18
refactor(schedule): prototype Effect driver coordination (blocked)

onmax:
BLOCKED — NON-MERGEABLE DIAGNOSTIC

This stacked draft tests whether Schedule driver installation and mutation coordination can move to Effect without changing public Promise APIs or runtime behavior. It depends on #676 and intentionally targets refactor/effect-schedule-process-driver.

The prototype is behaviorally green, but it fails ViteHub's Effect pay-rent rule. The baseline is 484 lines; the prototype is 608 lines, a net increase of 124. Effect can own fibers and shutdown, but exact Schedule semantics still require AsyncLocalStorage provenance plus a Queue/Deferred release protocol around Effect's Semaphore. That duplicates the coordination model instead of deleting it, so this must not merge.

**Why this is blocked**

Schedule allows driver reconciliation to wake a handler that awaits a Runtime Schedule mutation. That mutation is logically reentrant: it must write inside current serialized ownership, defer reconciliation until the active snapshot completes, release the awaiting wake to avoid deadlock, then surface failure and rollback in the correct causal order.

Effect v4 beta.99 provides Semaphore, Queue, Deferred, FiberSet, Scope, and ManagedRuntime, but no async-provenance-aware reentrant critical section with after-commit work. Semaphore alone deadlocks. Running wake fibers in parallel changes snapshot ordering and close semantics. This prototype rebuilds the missing ownership with AsyncLocalStorage and manual releases, producing the extra 124 lines.

**Invariant result**

- Public Promise API and behavioral equivalence: PASS.
- Reentrant handler mutation without deadlock: PASS, but only with AsyncLocalStorage provenance and a parallel release queue.
- Snapshot order, rollback plus reconcile, and close-during-handler mutation: PASS.
- One Effect-native ownership model with less machinery: FAIL.
- Report pay-rent requirement, measured as deletion or material simplification: FAIL.

**Evidence**

- Driver characterization: 29/29 tests pass.
- Full Schedule suite: 13 files and 185/185 tests pass, with no type or declaration errors.
- Schedule build and publint pass.
- Workspace lint exits 0; pre-existing warnings remain outside this diff.
- Public interfaces and Promise return shapes are unchanged.
- No catalog or lockfile edits.

**What would unblock this**

Effect would need an endorsed provenance-aware reentrant critical-section primitive with deferred-after-commit work, or ViteHub would need to change Schedule's contract so wake handlers cannot await mutations during reconciliation. The second option is a breaking semantic regression and should not be accepted merely to force Effect adoption.

EFFECT ADOPTION STATUS
  seam: Schedule driver installation and mutation coordination
  public API: unchanged
  boundary: installScheduleRuntime remains Promise-based; Effect and FiberFailure stay private
  typed failures: DriverBoundaryError preserves original driver, store, and rollback causes; no speculative public Schedule codes
  resources/fibers: one ManagedRuntime and scoped driver; FiberSets own operations, wakes, and waitUntil work; Deferred owns the close trigger
  deleted: none at module scale; manual promise sets and tail moved into Effect structures, but provenance and reentrant release coordination remain
  todos: 0 in implementation; blocked on the missing primitive or a design decision
  confidence: high in behavioral evidence, low that this design pays rent

Dependency: #676
Base: refactor/effect-schedule-process-driver
Head: refactor/effect-schedule-driver-coordinator
Status: BLOCKED — diagnostic only, do not merge.


onmax:
**Head:** `eb892f596371301402aaf2737241a636fc845b12`

**Verdict:** close

**Direction:** Preserve this as a diagnostic result and do not merge the prototype. The pull request's authoritative title and body explicitly mark it blocked and non-mergeable because the Effect conversion reproduces the existing coordination model instead of simplifying it.

**Evidence:**
- The exact-head diff adds 577 lines and deletes 453 lines across two Schedule driver files: a net increase of 124 lines.
- The body reports 29/29 driver characterization tests and 185/185 Schedule tests passing, so behavioral compatibility is established.
- The same evidence says reentrant mutation still requires AsyncLocalStorage provenance plus Queue/Deferred release coordination around Effect's Semaphore.
- Dependency PR #676 remains open and its branch has advanced from this PR's merge base `d27b66bf4af05354674d7f1a9e40b34629b8f381` to `5c0b0b7171cf6573b8e9fe95e0d37909b9ca7b6c`.

**ViteHub precedent:** Repository direction requires dependencies to pay rent and favors deletion, existing primitives, and the smallest obvious runtime contract. A private Effect boundary that adds machinery without changing the public Promise API or removing coordination complexity does not meet that bar.

**Strongest challenge:** The prototype is behaviorally green and moves fibers/shutdown into Effect-owned structures. That could justify merging as an architectural stepping stone, but the stated experiment specifically requires one Effect-native ownership model with less machinery; passing tests do not overcome the failed success criterion.

**Next action:** Close PR #678 without modifying or merging the diagnostic branch. Revisit only if Effect gains an endorsed provenance-aware reentrant critical-section primitive with after-commit work, or ViteHub intentionally changes Schedule's reentrant mutation semantics.

PR #677, opened 2026-07-18
refactor(agent): finalize invocations with Effect

onmax:
Agent Invocation finalization now has one Effect-cached completion that owns the first finish outcome and shares its result with every later caller. The public Promise, `Response`, and async-iterable APIs remain unchanged; close still precedes finish hooks, Workspace commits still occur only after successful finalization, and original failures retain identity unless a separately failing finish lifecycle requires the existing ordered `AggregateError`.

The public artifact guard rejects both `effect` and `effect/*` imports from every emitted Agent declaration. It also walks the built Cloudflare provider and browser-safe messages/output bundle graphs, so a future shared-chunk edge that pulls Effect into those artifacts fails the package suite.

Validation: the Agent build, publint, and typecheck pass. The restricted run passed 1,266 of 1,268 tests; its only failures were the two localhost listeners blocked by `listen EPERM`, and both affected files then passed 21/21 with localhost permission (1,287 passing test executions across the two runs).

EFFECT ADOPTION STATUS
  seam: Agent Invocation finalization
  decision: adopted
  public API: unchanged Promise, Response, and async-iterable contracts
  boundary: packages/agent/src/internal/effect-runtime.ts
  typed failures: AgentEffectFailure wraps original finish causes; Exit/Cause triage restores exact public identities and contains defects/interruption without FiberFailure
  resources/fibers: Deferred first outcome plus cached completion; no long-lived fibers
  deleted: manual close-once wrapper, finishFailedAgentInvocation ladder, finishLifecycleStarted boolean, and hand-written finish-failure aggregation
  artifacts: declarations and provider/browser bundle graphs remain Effect-free
  todos: 0
  confidence: high

DEPENDENCY STATUS
  #673: merged into the foundation branch
  #670: open at accepted final head 9331ce1c421ea0240c4c835e2b97332f43396c7e; this draft remains stacked on refactor/effect-error-contract from historical ancestor 4898061

PR #670, opened 2026-07-18
feat(runtime): establish error and Effect foundations

onmax:
## Summary

This establishes the shared boundary for ViteHub's selective Effect migration: one public structured-error base, Agent Capability cleanup owned by an Effect Scope, exact Effect 4 dependency ownership, and deterministic installed-consumer fixtures. Public Promise, Web, and ViteHub APIs remain Effect-free.

`ViteHubError` exposes stable codes, safe messages, allowlisted JSON details, truthful retry/request metadata, and a protected cause that never serializes. Its public shape is now a frozen construction-time snapshot: details are defensively cloned through a bounded accessor-free JSON validator, core fields are non-writable, and an own non-enumerable serializer cannot be shadowed after construction. Derived serializers are captured before sealing so Queue-style subclasses remain compatible, while structural borrowers retain first-call snapshot compatibility until their owning package closes construction itself. `CapabilityDeniedError` publishes a denial reason only when policy code explicitly marks it safe. Agent Capability cleanup uses one sequential Scope plus explicit Exit/Cause interpretation, preserving original and ordered aggregate cleanup failures without exposing `FiberFailure`.

The catalog pins every ViteHub-owned importer to `effect@4.0.0-beta.99`. Automated ownership and artifact tests require the source importers to equal `{ agent, schedule }`, recognize all TypeScript import forms, and reject Effect or `FiberFailure` from public declarations and browser/provider bundles. The existing optional UploadThing / `@effect/platform` Effect 3 subtree remains isolated and is covered by the same graph proof.

Fresh consumer, Deno, and Netlify fixtures set `msgpackr-extract: false`; installed-consumer contracts prove the JavaScript fallback and bundled Agent invocation so optional native acceleration cannot make generated-project validation host-dependent. The Runtime fixture now packs and installs the real package tarball before compiling, so its public error proof covers the published package rather than a copied monorepo directory. npm pack, npm install, and strict TypeScript compilation each have a hard 15-second child-process timeout. Pack and install complete in suite setup before compilation can start, and wider test envelopes allow a killed child to close before cleanup without adding retries.

## Integrated and corrective error work

[...]

PR #679, opened 2026-07-18
refactor(agent): centralize stream lifetimes with Effect

onmax:
Replaces the three duplicated Agent stream, Response body, and UI stream cleanup wrappers with one scoped Effect Stream boundary. The probe preserves lazy pull/backpressure, Response metadata, consumer cancellation reasons, AbortSignal reason identity, source `return()` exactly once, and ordered primary-plus-cleanup aggregation without exposing Effect or FiberFailure through the public API.

EFFECT ADOPTION STATUS

BLOCKED. Effect v4’s Web Stream adapters cannot preserve this contract directly: `Stream.toReadableStream` drops `cancel(reason)`, `Stream.fromReadableStream` cancels without the caller reason, and the async-iterable adapter does not initialize its source until the first pull. Preserving never-pulled cancellation and raw return/cancel error identity therefore requires a 174-line ViteHub bridge; the complete diff is 215 additions and 144 deletions, including 24 lines of new contract tests. That does not pass the migration report’s pay-rent gate, so this draft should not merge unless Effect gains the missing interop or the boundary can be reduced materially.

The focused capability-runtime and runtime suites pass, Agent typecheck passes, the Agent build and publint pass, and generated declarations contain no `effect` import or `FiberFailure`. Remote [CI run 29632376915](https://github.com/vite-hub/vitehub/actions/runs/29632376915) passed lint, typecheck, contracts, the tutorial fixture, provider verification for Vercel/Cloudflare/Netlify, consumer tests, and docs. Its sole failure was the unrelated `local-sandbox.test.ts` case “preserves roots owned by live processes”: 1,267 tests passed and that test lost `/tmp/vitehub-harness/owner-4243-…/keep.txt`. This branch changes only capability cleanup, the new stream-lifetime helper, stream-output cleanup, and their focused test, so no sandbox fix belongs here; the failing test uses a fixed fake PID under a temp directory shared by parallel workers and is susceptible to cross-worker reclamation.

DEPENDENCY STATUS

#677: open; this draft is stacked on `refactor/effect-agent-invocation-lifecycle` at `e20dcba`.

#670: open; inherited through #677.

### DIFF
--- a/packages/agent/src/capability-runtime.ts
+++ b/packages/agent/src/capability-runtime.ts
@@ -14,8 +14,12 @@ import {
   resolveInputAgentInvoker,
 } from "./invoker.ts"
 import { runObservedAgentHook } from "./hooks.ts"
-import { nextWithAbort } from "./internal/abortable-stream.ts"
 import { openAgentCapabilityScope } from "./internal/capability-scope.ts"
+import {
+  withAgentReadableStreamCleanup,
+  withAgentStreamCleanup,
+} from "./internal/stream-lifetime.ts"
+import type { AgentStreamCleanupOutcome } from "./internal/stream-lifetime.ts"
 import type {
   AgentCapabilitiesInput,
   AgentCapabilitiesResolverContext,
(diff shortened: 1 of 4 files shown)


onmax:
**Head:** `812b4116eab20a3e2d8e3cdd4a26e76d4cf67296`
**Verdict:** close

**Direction:** Do not merge this Effect stream-lifetime bridge. Close the draft because the authoritative body says it fails the adoption pay-rent gate, and current repository evidence offers no materially smaller implementation.

**Evidence:** Relative to `refactor/effect-agent-invocation-lifecycle` at `e20dcba`, the diff replaces 144 lines with 215 additions, including a new 174-line `stream-lifetime.ts` adapter. The body records that Effect v4 adapters lose caller cancellation reasons and never-pulled cancellation behavior, while the public API is unchanged. PR #677 demonstrates the accepted precedent: Effect owns a lifecycle only when it deletes manual machinery and preserves a compact, analyzable boundary.

**ViteHub precedent:** Repository guidance prefers deletion and existing primitives, permits extra internal machinery only when the obvious external contract justifies the cost, and rejects feature/abstraction creep. This bridge adds substantial private machinery without improving the external contract.

**Strongest challenge:** Centralizing three duplicated cleanup wrappers could reduce drift and consolidate subtle ordering rules. That benefit is outweighed here because the consolidation is larger than the removed code and must emulate semantics the chosen primitive does not support.

**Next action:** Close PR #679 without modifying or pushing the branch. Revisit as a new change only if Effect gains the required interop or repository evidence yields a materially smaller boundary.

PR #681, opened 2026-07-18
fix(shell): dispose background processes

onmax:
Shell Session now owns every background process until explicit stop or disposal. Provider stops are cached exactly once across synchronous throws and asynchronous rejections, explicit stop deregisters the process, and disposal atomically closes admission. It snapshots in-flight starts and registered processes, starts registered-process cleanup concurrently with pending provider startup, waits for late processes to stop before resolving, and returns every stop failure through AggregateError in deterministic registration order. Duplicate provider IDs remain independently owned, while the public Promise and observation contracts stay unchanged.

The two latest commits preserve that model while closing the remaining race: disposal now awaits provider starts already in flight, and registered stops are launched concurrently rather than serialized behind those starts. The Effect Scope prototype remains rejected under the migration report's pay-rent rule because it added +72/-17 lines to Session plus a 43-line interpreter boundary while retaining the process Map. The completed plain owner is production +71/-25 and deletes the ID-keyed Map for ownership Sets that handle pending starts and duplicate IDs without a second cleanup system.

Source accounting against #676: production +71/-25, tests +165/-42, total +236/-67.

At exact head d6783e376dca30e41c9adeed368666736c87e206, Shell build/publint, typecheck, focused Oxlint, and all 21 tests passed locally; all GitHub code and deployment checks later completed green. This pull request is now closed after dependency #676 merged and is not part of the active merge path.

DEPENDENCY STATUS
  base: #676 at d47c854576517a12fbfa75371b4328cf1bceb5f6
  merge: after #676, or retarget once #676 lands

EFFECT ADOPTION STATUS
  seam: Shell Session background-process lifetime
  public API: unchanged
  boundary: none; retained plain
  typed failures: AggregateError preserves every provider stop cause in registration order
  resources/fibers: process and pending-start ownership Sets; cached stop per process; cached dispose per session; no fibers
  deleted: ID-keyed process Map, marker-only disposal, repeated provider-stop execution, rejected Scope interpreter
  blocked: Effect Scope fails source pay-rent and duplicates ownership
  todos: 0
  confidence: high

> [!WARNING]
> **Babysitter status:** this pull request is closed and no longer has an active merge path.
>
> If this Shell change is still required, recover it in a fresh branch from the surviving target and revalidate there.

### DIFF
--- a/packages/shell/src/runtime/session.ts
+++ b/packages/shell/src/runtime/session.ts
@@ -106,7 +106,7 @@ class RuntimeShellSession implements ShellSession {
     })
     let stopTask: Promise<ShellObservation> | undefined
     let trackedProcess: ShellProcess
-    const stop = () => stopTask ??= process.stop().finally(() => {
+    const stop = () => stopTask ??= Promise.resolve().then(() => process.stop()).finally(() => {
       this.#processes.delete(trackedProcess)
     })
     trackedProcess = { ...process, stop }
--- a/packages/shell/src/runtime/session.ts
+++ b/packages/shell/src/runtime/session.ts
@@ -155,11 +155,14 @@ class RuntimeShellSession implements ShellSession {
   dispose() {
     this.#disposed = true
     return this.#disposeTask ??= (async () => {
-      const starts = await Promise.allSettled(this.#starts)
+      const pendingStarts = [...this.#starts]
       this.#starts.clear()
       const processes = [...this.#processes.values()]
       this.#processes.clear()
-      const results = await Promise.allSettled(processes.map(process => process.stop()))
+      const [starts, results] = await Promise.all([
+        Promise.allSettled(pendingStarts),
+        Promise.allSettled(processes.map(process => process.stop())),
+      ])
       const failures = [...starts, ...results]
         .flatMap(result => result.status === "rejected" ? [result.reason] : [])
       if (failures.length > 0) throw new AggregateError(failures, shellSessionStopFailureMessage)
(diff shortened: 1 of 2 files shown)


onmax:
**Head:** `d6783e376dca30e41c9adeed368666736c87e206`
**Verdict:** proceed

**Direction:** Keep the repair scoped to Shell Session ownership of provider-started background processes: stop each tracked process exactly once, deregister explicit stops, close admission before disposal snapshots tracked processes, preserve provider failure causes in registration order, and clean up a provider result that arrives after disposal.

**Evidence:** The title and body require disposal-owned background processes with unchanged public Promise and observation contracts. The exact diff changes only `packages/shell/src/runtime/session.ts` and focused runtime tests. Rebasing onto dependency #676 head `d47c854` preserved that diff; Shell build/publint, typecheck, focused Oxlint, and all 21 tests pass locally.

**ViteHub precedent:** The existing `ShellSession` is the host-independent runtime ownership boundary, and its existing `dispose()`, `listProcesses()`, and wrapped `ShellProcess.stop()` contracts are the obvious primitives to strengthen without adding a second cleanup system or framework-specific API.

**Strongest challenge:** Disposal can race an unresolved provider `startProcess` call, and a provider stop can throw before returning a Promise. The implementation tracks pending starts, launches registered stops concurrently, and establishes the cached stop Promise before invoking provider cleanup, so repeated explicit stop/dispose calls never invoke cleanup twice and duplicate provider IDs remain independently owned.

**Next action:** Wait for exact-head checks and Codex review. If dependency #676 merges, retarget to its merged base, reverify, and merge; otherwise retain the external dependency blocker.

onmax:
@codex review

onmax:
@codex review

onmax:
@codex review

onmax:
@codex review

PR #680, opened 2026-07-18
refactor(workspace): scope trusted-host sessions with Effect

onmax:
Keeps the trusted-host command protocol as the existing compact Promise state machine and gives Effect the ownership boundary it can improve: one parent Scope owns the materialized root, each active command registers one finalizer in a child Scope, and parent close drains those children before releasing the root. Close during execution still returns exit 130, timeout returns 124, the 100ms SIGKILL fallback remains, and stdout, stderr, environment, cwd, and spawn-error behavior stay unchanged.

The shared resource Scope replaces the duplicated sandbox Scope implementation, so sandbox and trusted-host use one lifecycle mechanism. Effect owns child registration, finalizer ordering, concurrent close, and exactly-once root release; an explicit wrapper-owned closed flag guards the public session contract without reading non-contract `scope.state` internals. The Promise still owns the Node child-process event protocol.

Third-party cleanup promises enter through `tryWorkspacePromise`. Each infallible Scope finalizer catches its typed `WorkspaceEffectFailure`, records the original cause in an Effect `Ref`, and completes without `orDie` or defect promotion. Only after `Scope.close` does the Promise boundary read those causes and merge them with operation or root-release failures in the existing order. Cleanup-only failures retain their original identity; operation plus cleanup failures return `AggregateError([operation, cleanup])`; parent close preserves child plus root cleanup ordering as `AggregateError([child, root])`; no path exposes `WorkspaceEffectFailure` or `FiberFailure`. Actual Effect defects stay defects and retain their exact identity.

The production delta from #675 is +280/-127, or +153 net lines; tests add +415/-8. That increase pays for one shared sandbox/trusted-host owner, typed cleanup capture, child-before-root ordering, closed-scope admission, finalizer-registration and abort races, exactly-once concurrent close, ordered preservation of child operation, child cleanup, and root cleanup failures, and direct regression coverage for the typed-failure versus defect boundary.

## Hard dependency

This PR targets `refactor/effect-workspace-sandbox-scope` at `e9216213` and must not merge before #675.

Proof at head `0a3ad8c`: exact-head CI passes, including the full Workspace suite at 30 files and 449 tests with no type errors; focused Effect runtime, trusted-host, and sandbox tests pass 61/61; docs, consumer, provider verification, preview package, and Workers checks pass. The external Vercel preview is rate-limited; its repository `verify (vercel)` check passes.

EFFECT ADOPTION STATUS
[...]

PR #682, opened 2026-07-18
fix(workspace): cancel source HTTP requests

onmax:
Propagates Workspace materialization cancellation into the shared HTTP request boundary, where `fetch` can actually stop the work. Caller cancellation is composed with request timeouts, preserves the caller's exact abort reason, and prevents the GET/HEAD retry path from starting another attempt after cancellation; ordinary transport failures still retry as before.

This is intentionally a plain TypeScript prerequisite rather than an Effect adoption: interruption would be false unless the underlying operation receives the signal. It keeps the later Effect migration honest while changing no successful-request API behavior.

**Hard dependency:** stacked on #680 and must be rebased after that PR lands.

```text
EFFECT ADOPTION STATUS
  seam: Workspace fetch-source request cancellation prerequisite; no Effect implementation added
  public API: unchanged; internal HttpRequestOptions gains an optional AbortSignal
  boundary: executeHttpRequest remains the single plain Promise interpreter and passes the composed signal to fetch
  typed failures: existing request failures remain unchanged; caller cancellation preserves the raw AbortSignal.reason identity
  resources/fibers: none
  deleted: none; retained the existing request interpreter and non-cancellation retry policy
  todos: 0
  confidence: high
  dependency: #680
```

### DIFF
--- a/packages/internal/src/http-request.ts
+++ b/packages/internal/src/http-request.ts
@@ -18,6 +18,7 @@ export interface StandardSchemaV1<T = unknown> {
 }
 
 export interface HttpRequestOptions {
+  abortSignal?: AbortSignal
   body?: unknown
   cookies?: Record<string, string>
   headers?: Record<string, string>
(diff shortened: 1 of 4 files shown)


onmax:
**Head:** `8735d399bcb931cf3d05bb0af29cfdeebdf27375`
**Verdict:** proceed
**Direction:** Propagate Workspace Source cancellation to the shared HTTP request boundary, compose it with the existing timeout signal, preserve the caller abort reason throughout fetch and response decoding, and prevent cancellation from entering the GET/HEAD retry path.
**Evidence:** The exact diff adds the existing ViteHub `abortSignal` convention to `HttpRequestOptions`, supplies each Source context signal through the shared fetch-source request builder, and covers in-flight fetch cancellation, response-decoding cancellation, pre-aborted callers, timeout composition, and ordinary retry behavior. Successful-request behavior and the public Workspace API remain unchanged.
**ViteHub precedent:** Source, Agent Invocation, harness, and Workspace server primitives consistently expose optional `AbortSignal` values and pass them to the operation that owns the underlying work; `executeHttpRequest` is the existing shared fetch interpreter.
**Strongest challenge:** A caller can abort after fetch resolves but while response decoding is pending; the interpreter now reasserts the caller signal around decoding/schema work so the exact caller reason wins over a generic transport `AbortError`.
**Next action:** Wait for exact-head checks, perform a fresh exact-diff audit, then request the single exact-head Codex review.

onmax:
@codex review

PR #683, opened 2026-07-18
refactor(sandbox): prototype Effect attempt ownership (blocked)

onmax:
> [!CAUTION]
> **BLOCKED — DO NOT MERGE.** This draft preserves the tested Effect experiment required by the migration process, but the experiment fails ViteHub's pay-rent rule.

## EFFECT ADOPTION STATUS

**RETAIN PLAIN TYPESCRIPT.**

The Effect version replaces both Cloudflare retry loops and proves the intended invariants: exact `[1s, 2s, 5s, 10s, 15s]` delays, one deadline per attempt, cleanup before retry, one stop per acquired client, ordered execution/cleanup causes, and raw Promise errors without `FiberFailure`.

An equivalent shared async TypeScript helper expresses the same runtime contract in 112 production lines. This Effect helper takes 134 production lines, so the complete slice is **+70 net lines with Effect versus +48 net lines in plain TypeScript**: 22 lines, or about 31%, worse on net source cost, plus a new runtime dependency and interpreter boundary.

The seam also has no outer cancellation signal to propagate into the Cloudflare SDK calls. Effect's interruption and uninterruptible release machinery therefore cannot materially strengthen the runtime contract here; deterministic `TestClock` coverage alone is insufficient pay rent for the added machinery.

## Validation

- Focused attempt-policy tests: 5/5 passed.
- Full `@vite-hub/sandbox` suite: 55/55 passed across 14 files.
- Sandbox typecheck passed again after the final dependency rebase.
- Sandbox build and declaration generation passed.
- Publint passed.
- Built public declarations contain no Effect or `FiberFailure` leakage.
- Frozen install passes on the rebased graph; the lockfile diff contains only the three-line `packages/sandbox` Effect importer entry.

This blocked experiment is rebased onto final #676 at `9ad2a81b3169869407e0099d765dc9de5ba5f22f`, which in turn is based on the repaired #670 Effect foundation. It should remain closed and unmerged unless the boundary gains real cancellation or another correctness property that plain TypeScript cannot express with the same clarity.

### DIFF
--- a/packages/sandbox/package.json
+++ b/packages/sandbox/package.json
@@ -81,6 +81,7 @@
     "test": "vp run -t @vite-hub/sandbox#build && vp test"
   },
   "dependencies": {
+    "effect": "catalog:effect",
     "esbuild": "catalog:tooling",
     "h3": "catalog:unjs",
     "local-pkg": "catalog:tooling",
--- a/packages/sandbox/src/internal/shared/cloudflare-attempt.ts
+++ b/packages/sandbox/src/internal/shared/cloudflare-attempt.ts
@@ -1,0 +1,134 @@
+import { Cause, Data, Effect, Exit, Schedule } from 'effect'
+
+import { SandboxError } from '../../sandbox/errors'
+import { CLOUDFLARE_SANDBOX_RETRY_DELAYS_MS } from './cloudflare-retry'
+
+class CloudflareAttemptFailure extends Data.TaggedError('CloudflareAttemptFailure')<{
+  readonly cause: unknown
+  readonly retryable: boolean
+}> {}
+
+interface CloudflareOperationOptions<T> {
+  isRetriable: (error: SandboxError) => boolean
+  mapError: (error: unknown) => SandboxError
+  operation: string
+  run: () => Promise<T>
+  timeout: number
+}
+
+interface CloudflareSandboxExecutionOptions<TResource, TResult> {
+  acquire: () => Promise<TResource>
+  isRetriable: (error: SandboxError) => boolean
+  mapError: (error: unknown) => SandboxError
+  release: (resource: TResource) => Promise<void>
+  use: (resource: TResource) => Promise<TResult>
+}
+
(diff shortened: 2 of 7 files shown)


onmax:
**Head:** `68b0fb5662d26cb9335d722b9baebb1c4f4feb2a`

**Verdict:** close

**Direction:** Retain the existing plain TypeScript Cloudflare attempt ownership and close this deliberately blocked Effect experiment without merging it.

**Evidence:** The authoritative title marks the prototype blocked. The body explicitly says “BLOCKED — DO NOT MERGE” and “RETAIN PLAIN TYPESCRIPT,” and records that the Effect implementation is 22 net lines (about 31%) costlier than the equivalent shared TypeScript helper, adds a runtime dependency/interpreter boundary, and cannot improve cancellation because the boundary has no cancellation signal to propagate. CI and focused tests establish that the experiment works, but do not overturn its failed adoption criterion. The stacked base PR #676 remains open, so this PR also is not independently mergeable into its intended final base.

**ViteHub precedent:** Repository guidance says to prefer existing primitives, avoid feature creep and accidental plumbing, and make dependencies pay rent at clear runtime boundaries. Effect is justified in #676 for schedule process wakes; that does not establish that every retry loop should cross an Effect interpreter boundary. Keeping a smaller plain TypeScript helper at this non-cancellable SDK seam follows the project’s composability and obvious-API rules.

**Strongest challenge:** The prototype does consolidate duplicate retry loops and provides excellent deterministic `TestClock` coverage of delay, deadline, cleanup, and error invariants. If this seam later gains a real cancellation signal or another correctness property that Effect uniquely enforces, these benefits could justify revisiting adoption.

**Next action:** Close PR #683 as a completed negative experiment; do not merge or modify its implementation.

onmax:
Closing this completed prototype per its recorded adoption verdict: the Effect boundary does not pay rent at this non-cancellable seam, so ViteHub should retain the plain TypeScript design. The direction evidence is cached above for future reconsideration if the boundary gains cancellation or another uniquely enforceable property.

PR #684, opened 2026-07-18
fix(workflow): own OpenWorkflow backend lifecycle

onmax:
OpenWorkflow runtime caching now has one generation owner. Concurrent gets for the same storage config share one acquisition, rejected initialization is evicted, and reset atomically swaps the cache before waiting so new gets cannot join a generation being closed. A pending pre-reset get rejects with the stable `OPENWORKFLOW_RUNTIME_RESET` WorkflowError after its backend is closed, while post-reset gets use the new generation.

Reset closes every fulfilled backend exactly once and preserves cleanup failures in cache order: one failure rejects as `WorkflowError(OPENWORKFLOW_BACKEND_CLOSE_FAILED)`, while multiple failures reject as an AggregateError containing each WorkflowError and original cause. The importer is captured per acquisition so reset cannot mix importer generations. Public Promise declarations stay unchanged, and OpenWorkflow's durable SDK retry mapping is untouched.

Effect Scope is intentionally rejected under the migration report's pay-rent rule. The correct plain owner is +46/-20 production lines; Scope would leave the same Map, acquisition Promise, generation swap, rejection eviction, and reset/get admission checks, then add an interpreter and Cause mapper merely to replace `backend.stop()`. This PR is the retained-plain outcome, with no Effect dependency or lockfile change.

Proof: Workflow typecheck passes; the package build and publint pass; the focused runtime file passes all 53 tests and the complete Workflow suite passes all 92 tests across five files; the full workspace build, focused Oxlint, and repository Fallow pass; generated declarations and runtime output contain no Effect or FiberFailure leakage.

DEPENDENCY STATUS
  base: #676 at 5c0b0b7171cf6573b8e9fe95e0d37909b9ca7b6c
  merge: after #676, or retarget once #676 lands

EFFECT ADOPTION STATUS
  seam: cached OpenWorkflow backend lifecycle
  public API: Promise signatures unchanged; reset failures now use stable WorkflowError codes
  boundary: none; retained plain
  typed failures: OPENWORKFLOW_RUNTIME_RESET; OPENWORKFLOW_BACKEND_CLOSE_FAILED; AggregateError preserving ordered cleanup causes
  resources/fibers: one Map generation owner; reset swaps then closes the detached generation; no fibers
  deleted: sticky rejected acquisitions, unsnapshotted reset race, first-failure-only cleanup, live importer lookup during acquisition
  blocked: Scope cannot own cache identity, deduplication, generation, or reset admission and fails source pay-rent
  todos: 0
  confidence: high

### DIFF
--- a/packages/workflow/src/runtime/openworkflow.ts
+++ b/packages/workflow/src/runtime/openworkflow.ts
@@ -1,4 +1,5 @@
 import { runWorkflowHandler } from "./execute.ts"
+import { WorkflowError } from "../errors.ts"
 
 import type { RetryPolicy } from "openworkflow"
 import type { ResolvedWorkflowOptions, WorkflowDefinition, WorkflowDeferOptions, WorkflowProviderStep, WorkflowRun, WorkflowRunStatus, WorkflowRuntimeConfigValue, WorkflowRuntimeEnvDeclarationLike, WorkflowStepOptions } from "../types.ts"
@@ -24,7 +25,7 @@ interface OpenWorkflowRuntime {
   workflows: Map<string, OpenWorkflowRunnable>
 }
 
-const runtimes = new Map<string, Promise<OpenWorkflowRuntime>>()
+let runtimes = new Map<string, Promise<OpenWorkflowRuntime>>()
 
 function readEnv(name: string): string | undefined {
   const env = (globalThis as { process?: { env?: Record<string, string | undefined> } }).process?.env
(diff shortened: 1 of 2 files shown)


onmax:
**Head:** aa3b12dbaddfb86054805d70eca95e3a671cdeea
**Verdict:** proceed

**Direction:** Keep the repair scoped to the cached OpenWorkflow backend lifecycle: one acquisition per storage key and generation, rejected-acquisition eviction, atomic generation detachment on reset, deterministic close semantics, and stable ViteHub workflow errors. Preserve the existing Promise API and avoid adding an Effect dependency where it does not own the cache lifecycle.

**Evidence:** The title and body ask for lifecycle ownership; the two-file diff implements only that seam and adds focused concurrency, retry, reset-race, close-once, and ordered-failure tests. The branch is exactly one commit on its declared base (#676 at 5c0b0b7), is cleanly mergeable, and its full GitHub CI is green.

**ViteHub precedent:** Existing Workflow runtime boundaries expose provider-neutral Promise APIs and use WorkflowError with stable provider/code metadata. The project guidance prefers small host-independent primitives, explicit lifecycle contracts, existing primitives, and dependencies that pay rent.

**Strongest challenge:** The new owner must not admit a pre-reset acquisition into the new generation or lose/duplicate backend cleanup, and rejected initialization must not poison the key. The base is still the open dependency PR #676, so this PR must remain targeted to that exact base until the dependency lands.

**Next action:** Review the exact diff against the base, run focused verification, repair any actionable finding, then request exact-head Codex review and merge only after its terminal positive result and green checks.

onmax:
@codex review

PR #685, opened 2026-07-18
fix(workflow): own OpenWorkflow worker shutdown

onmax:
Makes a successfully started OpenWorkflow worker own its AbortSignal and process-signal listeners. Public stop removes every listener before calling the SDK, concurrent and repeated stops share one cached result, and provider failures reject with `WorkflowError` code `OPENWORKFLOW_WORKER_STOP_FAILED`.

Abort and process callbacks stay synchronous because their hosts do not await returned Promises; stop failures are routed through the optional `onError` hook, with a process warning/console fallback, instead of becoming unhandled rejections. `createOpenWorkflowWorker` remains caller-owned and unchanged.

EFFECT ADOPTION STATUS: retained plain. Effect Scope would still require the same public stop cache, returned-worker method replacement, listener registration state, and synchronous host error bridge, so it adds an interpreter boundary without deleting ownership machinery.

Depends on #676, which already contains the canonical OpenWorkflow backend lifecycle from #684.

Validation: Workflow build, publint, typecheck, focused lint, and the full 95-test package suite pass.

### DIFF
--- a/packages/workflow/src/runtime/openworkflow-worker.ts
+++ b/packages/workflow/src/runtime/openworkflow-worker.ts
@@ -1,5 +1,6 @@
 import { getOpenWorkflowRuntime, registerOpenWorkflowDefinition } from "./openworkflow.ts"
 import { getInlineWorkflowDefinitions, getWorkflowRuntimeConfig, getWorkflowRuntimeRegistry, setWorkflowRuntimeConfig, setWorkflowRuntimeRegistry, takeInlineWorkflowDefinitionForModule } from "./state.ts"
+import { WorkflowError } from "../errors.ts"
 
 import type { ResolvedWorkflowOptions, WorkflowDefinition, WorkflowDefinitionRegistry } from "../types.ts"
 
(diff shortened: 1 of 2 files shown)


onmax:
**Head:** c673029fd46b405bb77d385f0e04dd7051793f11
**Verdict:** proceed

**Direction:** Make startOpenWorkflowWorker own shutdown resources after a successful start: remove registered abort/process listeners before stopping, coalesce repeated/concurrent public stops, translate provider stop failures to WorkflowError, and safely report failures from synchronous host callbacks. Keep createOpenWorkflowWorker caller-owned.

**Evidence:** The PR title/body and two-file diff agree; the branch is a single commit directly atop the current base; CI is green; tests cover cached public stop, listener cleanup, abort-triggered failure reporting, and process-triggered synchronous callbacks.

**ViteHub precedent:** The base includes #684's canonical OpenWorkflow backend lifecycle and uses provider-specific WorkflowError codes at runtime boundaries. Lifecycle ownership belongs in the high-level start primitive, not the low-level create primitive.

**Strongest challenge:** Replacing the returned SDK worker stop method and adding onError are observable API choices. They are justified only if public stop must share the same lifecycle cleanup/cache and host callbacks need an explicit failure bridge; documentation/export precedent still needs confirmation.

**Next action:** Review package precedent and the exact base diff, run focused verification, minimize any accidental complexity, then perform an independent base-branch code review before requesting exact-head Codex review.

onmax:
@codex review

PR #703, opened 2026-07-18
feat(blob): support served response headers

onmax:
Adds validated static `headers` to `blob.serve` and serializes them into the generated Nitro handler. The handler applies them before Blob storage writes its response metadata, so `Content-Type`, `Content-Length`, and `ETag` remain authoritative.

This lets applications configure cache and security headers on the established generated serving route instead of maintaining a parallel source handler.

Covered by Blob config, public type, generated-handler, build, package, README, and server-primitives documentation tests.

### DIFF
--- a/docs/content/docs/server-primitives/blob.md
+++ b/docs/content/docs/server-primitives/blob.md
@@ -77,7 +77,7 @@ export default defineConfig({
 | `blob: { stores: Record<string, BlobStoreConfig> }` | Defines named Blob Stores. `stores.default` is required. |
 | `blob: { serve: false }` | Disables Blob route generation. This is the default. |
 | `blob: { serve: true }` | Generates an opt-in Nitro route at `/api/_vitehub/blob/**` for serving the Default Blob Store. |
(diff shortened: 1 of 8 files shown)


onmax:
**Head:** `172216fc5353d5839b9f644df2653d5ead0c255e`

**Verdict:** proceed

**Direction:** Extend the existing generated Blob serving route with validated static response headers; keep Blob storage metadata authoritative for content headers and keep error responses free of the configured asset policy.

**Evidence:** The PR validates and serializes `serve.headers`, applies them before successful Blob metadata writes, and removes them when serving throws so cacheable asset policy cannot leak onto 404s. It covers public/resolved types, config, generated handling, package README, and public docs. Focused Blob tests passed 49/49, Blob typecheck passed, and docs tests passed 38/38.

**ViteHub precedent:** `blob.serve` already owns route generation, store selection, and public URL configuration. Static route response policy fits that primitive better than requiring an application-maintained parallel handler.

**Strongest challenge:** Applying cache headers before storage lookup could cache missing-Blob errors. The generated handler now removes every configured header before rethrowing a serving error while retaining Blob metadata precedence on success.

**Next action:** Wait for exact-head CI and Codex review, address any actionable findings, then merge.

onmax:
@codex review

onmax:
@codex review

PR #714, opened 2026-07-18
fix(workflow): close public error boundaries

onmax:
## Summary

- closes `WorkflowError` to ViteHub-owned codes, fixed code-derived messages, and code-specific provider, operation, name, and status details
- moves application-owned codes to the explicit `ApplicationWorkflowError` boundary, which validates bounded codes and messages and defensively clones JSON-safe plain-object details with depth and entry limits
- rejects cycles, BigInt, class instances, accessors, hostile proxies, reserved codes, and malformed SCREAMING_SNAKE_CASE codes without echoing raw diagnostics
- freezes trusted details and public fields while preserving subclass field initialization
- installs the standalone trusted serializer only when Runtime has not already installed its final sealed serializer, so the branch composes with #670 without redefining a non-configurable property

## Public contract and migration

This deliberately breaks the generic constructor introduced by #694.

```ts
// Before: implicitly publishes every message and detail supplied by the caller.
new WorkflowError({ code: "TRANSCRIPTION_FAILED", message: "Transcription failed.", details })

// After: explicit application-owned publication boundary.
new ApplicationWorkflowError({ code: "TRANSCRIPTION_FAILED", message: "Transcription failed.", details })

// ViteHub-owned failures no longer accept a caller message.
new WorkflowError({ code: "WORKFLOW_DISABLED" })
```

`ApplicationWorkflowError` deliberately publishes its validated `message` and cloned `details`; raw provider diagnostics belong in `cause`, which remains available on the error instance but is excluded from JSON. Application codes are non-reserved SCREAMING_SNAKE_CASE values up to 64 characters, messages are 1–512 characters, and details allow at most five levels and 64 total entries.

## Validation

- Workflow typecheck — passed
- Workflow build and publint — passed
- current-base focused source, installed-consumer, built-artifact, provider-operation, runtime, hostile-input, mutation, and subclass-field suite: 4 files, 85 tests — passed
- exact no-conflict merge with final #670 `4332ff188b68748e435d446aae7ebc1d1aa094fc` — passed
- merged-foundation Runtime and Workflow build/publint, Workflow typecheck, and the same 4-file/85-test suite — passed
- source and built subclass-with-own-field regressions were not included in the merged head; tests-only follow-up #715 preserves them
- standalone trusted serializer remains pinned against later prototype instrumentation — passed
- oxlint on the compatibility correction — passed
- `git diff --check` — passed
[...]

PR #694, opened 2026-07-18
feat(workflow): add structured error contract

onmax:
Builds `WorkflowError` on the shared `ViteHubError` foundation with one required-code object constructor and JSON-safe details. Built-in codes and constructor options are public, custom application codes remain supported, and `cause` stays available to protected diagnostics without crossing `toJSON()`. Unsafe workflow names, caller run IDs, provider URLs, response bodies, and provider messages remain internal.

Cloudflare `get`/`create`/`status`, Vercel `getRun`/`listSteps`/`start`/`cancel`/`resumeHook`, and OpenWorkflow import/connect/run/get failures cross one provider-operation mapper. Operational failures reject with `WORKFLOW_PROVIDER_OPERATION_FAILED`, the fixed message `Workflow provider operation failed.`, allowlisted `{ provider, operation, status? }` details, and the exact non-serialized cause. Existing `WorkflowError` and abort errors preserve exact identity.

Vercel results are decoded inside that mapper before reaching ViteHub callers. Malformed run identity, dates, status, step records, start IDs, and resume IDs reject through the stable provider failure contract; hostile result getters preserve their exact private cause. OpenWorkflow client construction and provider-result normalization likewise stay inside the corresponding connect/run/get operations, so malformed SDK results cannot escape as raw defects after a provider call resolves.

OpenWorkflow configuration validation remains programmer-facing before the provider boundary. Published declarations and JavaScript are guarded against Effect imports, `FiberFailure`, or an Effect dependency; this pure Promise mapper does not earn an Effect runtime.

This is an intentional breaking change. Migrate `new WorkflowError(message, options)` to `new WorkflowError({ code, message, cause?, details? })`, read provider context from `error.details?.provider`, and map HTTP responses at the server-route seam instead of reading the removed `error.httpStatus`.

The exact feature diff from historical #670 ancestor `6b986d55` is +275/-106 production lines, +618/-4 tests, +42/-0 installed-consumer fixtures, +47/-1 documentation, and +4/-0 package/lock metadata, for +986/-111 overall. The added code pays for the stable Workflow error vocabulary, one provider-operation mapper, safe public detail normalization, complete provider result decoding, constructor migration proof, redaction and exact-identity tests, and declaration/runtime artifact guards.

## Hard dependency

This draft targets `refactor/effect-error-contract`, contains historical #670 ancestor `6b986d55`, and must not merge before current foundation head `b9b334b70b2ccf049b39b5507783a05aafb2c667`.

[...]

PR #715, opened 2026-07-19
test(workflow): cover derived error serialization

onmax:
#714 merged before its final compatibility regressions reached the pull request. The current #694 branch correctly removes Workflow's redundant serializer and relies on the Runtime foundation's sealed construction-time snapshot, so this follow-up carries only the two missing regressions and adds no production serializer code.

The source test proves a consumer subclass can initialize its own field after `WorkflowError` construction while inheriting the trusted, non-writable serializer. The built-artifact test proves the same behavior from the published JavaScript boundary, including the exact safe JSON shape.

## Hard dependency

This draft targets `feat/workflow-error-contract` at exact base `6a980f6f0bbb9be3c979c01dfc2e3e19a0f9ba16` and must merge after #694. If that base advances, revalidate these tests against its final Runtime foundation before merging.

At exact head `bbf8881b`, Workflow build, typecheck, focused source and built-artifact tests, changed-file Oxlint, and `git diff --check` pass.

## EFFECT ADOPTION STATUS

- Seam: Workflow public error serialization compatibility.
- Public API: unchanged; this is tests only.
- Boundary: source and built `WorkflowError` subclass behavior.
- Typed failures: unchanged closed Workflow and application error contracts.
- Resources/fibers: none.
- Deleted: no production code is added; the Runtime-owned serializer remains the single implementation.
- Retained: derived consumer fields and sealed safe JSON serialization.
- TODOs: merge after #694 and revalidate if its base advances.
- Confidence: high at the exact head.

> [!WARNING]
> **Babysitter is blocked:** prerequisite pull request #694 is still open.
>
> Merge #694 into its target branch, then revalidate and merge this tests-only follow-up.

### DIFF
--- a/packages/workflow/test/errors.test.ts
+++ b/packages/workflow/test/errors.test.ts
@@ -65,6 +65,21 @@ describe("WorkflowError", () => {
     expect(() => new WorkflowError(options as never)).not.toThrow(secret)
   })
 
+  it("keeps subclass fields compatible with trusted serialization", () => {
+    class CustomWorkflowError extends WorkflowError<"WORKFLOW_DISABLED"> {
+      readonly metadata = "consumer-owned"
+    }
+
+    const error = new CustomWorkflowError({ code: "WORKFLOW_DISABLED" })
+    expect(error.metadata).toBe("consumer-owned")
+    expect(error.toJSON()).toEqual({ code: "WORKFLOW_DISABLED", message: "Workflow is disabled." })
+    expect(Object.getOwnPropertyDescriptor(error, "toJSON")).toMatchObject({
+      configurable: false,
+      enumerable: false,
+      writable: false,
+    })
+  })
+
   it("captures trusted serialization before prototype instrumentation", () => {
     const error = new WorkflowError({ code: "WORKFLOW_DISABLED" })
     const original = ViteHubError.prototype.toJSON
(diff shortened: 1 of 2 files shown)


onmax:
**Head:** `bbf8881b529eddd02616d89f7ec81599bae14f80`

**Verdict:** proceed

**Direction:** Keep this a tests-only follow-up covering consumer subclasses of `WorkflowError` at both the source and published-JavaScript boundaries. Do not add a Workflow-owned serializer.

**Evidence:** The diff adds only two focused regressions. Both prove that a derived class can initialize its own field after `WorkflowError` construction while retaining the trusted, non-configurable, non-enumerable, non-writable `toJSON` snapshot and exact safe JSON output. PR #694 is still open at `6a980f6f0bbb9be3c979c01dfc2e3e19a0f9ba16`, which is also this pull request's unchanged base head.

**ViteHub precedent:** The current base centralizes sealed construction-time error serialization in the Runtime foundation; Workflow inherits that primitive rather than duplicating serialization logic.

**Strongest challenge:** These regressions depend on the not-yet-merged Runtime/Workflow error contract in #694. Merging them before #694 reaches its target would either be impossible or detach the tests from the production behavior they specify.

**Next action:** Wait for #694 to merge, then reconcile this branch with its resulting target/base, rerun focused and repository verification, perform exact-head review, and merge if green.

PR #722, opened 2026-07-19
fix(vite-hub): resolve provider facade aliases

onmax:
Provider builds can redirect a ViteHub feature facade to a generated runtime implementation, but Vite may rewrite the `vite-hub/*` specifier to the framework distribution path before the dependency resolver sees it. The resolver now recognizes both forms and preserves the generated provider facade, which keeps discovered Sandbox definitions in the provider registry.

Adds focused contract coverage for direct and Vite-rewritten facade resolution. This can merge independently of #720, though both branches touch `frameworkDependencyResolver` and the later branch may need a small rebase.

### DIFF
--- a/packages/vite-hub/src/index.ts
+++ b/packages/vite-hub/src/index.ts
@@ -140,8 +140,11 @@ function frameworkDependencyResolver(
       )
     },
     resolveId(id, importer) {
-      const providerImportAlias = providerImportAliases[id]
-        ?? Object.entries(providerImportAliases).find(([specifier]) => frameworkProviderImportAliases[specifier] === id)?.[1]
+      const providerFacadeAliases = Object.entries(providerImportAliases)
+        .filter(([specifier]) => frameworkProviderImportAliases[specifier])
+      const providerFacadeTargets = new Set(providerFacadeAliases.map(([, target]) => target))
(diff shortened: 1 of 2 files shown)


onmax:
**Head:** `fd99bc83518692fb24178ab8ab6b7a651792dfaa`
**Verdict:** proceed
**Direction:** Preserve generated provider facades when Vite has already rewritten a public `vite-hub/*` feature specifier to its framework distribution target, without exposing provider-only optional dependency fallbacks to ordinary application imports.
**Evidence:** The focused resolver test passes all 28 tests with no type errors; edited files lint cleanly; the simplified repair now derives eligible aliases from the authoritative public framework facade map and regression-tests the Upstash case.
**ViteHub precedent:** `frameworkProviderImportAliases` is the authoritative public-facade mapping, while `configureProviderOptionalImportAliases` deliberately keeps the Upstash fallback out of the global alias map.
**Strongest challenge:** Looking up every `providerImportAliases` entry before importer checks would leak `@vite-hub/kv/runtime/upstash-driver` into normal application resolution; the pushed head excludes entries that are not paired with a public framework facade.
**Next action:** Await clean-environment CI and exact-head Codex review, address any actionable result, then merge.

onmax:
@codex review

onmax:
@codex review

PR #726, opened 2026-07-19
fix(blob): restore content type detection

onmax:
Restores `detectContentType` as an explicit `@vite-hub/blob/content-type` API and forwards it through `vite-hub/blob/content-type`. The helper previously landed through a stacked base but is absent from current `main`, leaving consumers to duplicate common image and PDF signature detection.

Adds the package exports, build entries, and provider-output aliases required for the subpath to work in published packages and bundled Cloudflare and Vercel outputs. Detection remains signature classification only, so callers still own Blob `contentType` metadata and complete-file validation.

### DIFF
--- a/docs/content/docs/reference/import-paths.md
+++ b/docs/content/docs/reference/import-paths.md
@@ -29,6 +29,7 @@ composition and explicit feature subpaths for application APIs.
 | `vite-hub/auth` and `vite-hub/auth/server` | Auth Definitions and server runtime helpers. |
 | `vite-hub/auth/agent` | Better Auth session mapping into Agent Invokers. |
 | `vite-hub/blob` | Blob Runtime Helpers and Blob Store access. |
+| `vite-hub/blob/content-type` | Detect common image and PDF signatures from leading bytes before upload. |
 | `vite-hub/box` | Box Definitions and trusted-host execution contracts. |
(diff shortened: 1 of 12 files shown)


onmax:
**Head:** `0f242424df7ad4ef88913965a6efcc1b9aee4fb3`
**Verdict:** proceed
**Direction:** Restore `detectContentType` as an explicit Blob subpath API in both the owner and framework packages, including provider-output resolution, tests, and concise safety documentation.
**Evidence:** The PR is one focused commit directly atop current `main`; `main` has no `detectContentType`; the title/body and 12-file diff consistently restore the missing API; CI, package preview, and all three provider verification jobs pass.
**ViteHub precedent:** Public server primitives use explicit owner-package subpaths plus `vite-hub` facade subpaths, declared build entries/exports, import-path documentation, package-contract tests, and provider aliases where bundling crosses runtime boundaries.
**Strongest challenge:** Signature matching can be mistaken for file validation or automatic Blob metadata inference. The implementation stays synchronous and classification-only, while docs explicitly keep metadata and complete-file validation with the caller.
**Next action:** Review the exact diff against `main`, run focused verification, obtain a terminal exact-head Codex review, then merge if no actionable findings remain.

onmax:
@codex review

PR #736, opened 2026-07-20
fix(agent): tighten default chat titles

onmax:
Replaces `title()`’s default summarization prompt with a compact topic-label contract that keeps the message’s language, key names, numbers, and identifiers while treating instruction-like message content as source text. The configured fallback remains part of the prompt, and custom templates and title delivery are unchanged.

Keeps the documented 80-character global default because callers already use `maxLength` for stricter platform limits, and that limit remains enforced after generation.

### DIFF
--- a/packages/agent/src/capabilities/title.ts
+++ b/packages/agent/src/capabilities/title.ts
@@ -75,17 +75,11 @@ export interface TitleOptions<TRuntimeConfig extends AgentRuntimeConfig = AgentR
 }
 
 const defaultTitleTemplate = [
-  "Summarize the user's request as a short title.",
-  "Return only the title.",
-  "Use 4-8 words when possible.",
-  `Keep it under {{ maxLength }} characters.`,
-  "Do not answer the request.",
-  "Use the user's language.",
-  "Do not use emoji.",
-  "Ignore chat platform mention/channel markup, bot names, and user IDs.",
-  `Use "{{ fallback }}" when the message is too vague.`,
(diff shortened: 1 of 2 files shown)


onmax:
**Head:** e611059590636d5cf3318896dac0490520aaa98b
**Verdict:** proceed
**Direction:** Replace the verbose default title() summarization prompt with a compact, injection-resistant topic-label contract while preserving language, salient literals, fallback behavior, and the existing enforced 80-character default.
**Evidence:** The PR title/body and two-file diff agree: only the built-in prompt and its focused runtime assertions change. Custom templates, delivery, cleanup, and maxLength enforcement remain untouched. The branch is mergeable with current main.
**ViteHub precedent:** title() already centralizes provider-agnostic prompt rendering in packages/agent/src/capabilities/title.ts, strips chat entity markup before rendering, and enforces maxLength after generation; the public docs describe outcomes and options rather than promising the old prompt wording.
**Strongest challenge:** The new prompt removes explicit “under maxLength,” “no emoji,” and mention-markup instructions. This is coherent because length is deterministically enforced after generation, markup is stripped before the model sees it, and “neutral words” narrows decorative output; focused tests must still prove instruction-like content is treated as source text and configured fallback is rendered.
**Next action:** Run the focused package verification and review the exact diff against main; repair only actionable correctness or documentation gaps before requesting exact-head Codex review.

onmax:
@codex review

PR #737, opened 2026-07-20
refactor(cli): reuse shared contributor contract

onmax:
Reuses the single ViteHub CLI contributor contract and namespace collector from `@vite-hub/internal/cli` across the public CLI and Database package. This removes the duplicate contract declarations and collector implementation so contributor ordering and feature replacement semantics have one owner.

There is no public API or behavior change; the CLI build continues to inline the private internal declarations into `@vite-hub/cli`.

### DIFF
--- a/packages/cli/src/index.ts
+++ b/packages/cli/src/index.ts
@@ -10,7 +10,30 @@ import { resolve } from "pathe"
 import { runProvision } from "./provision.ts"
 
 import type { InlineConfig, ResolvedConfig } from "vite"
-import type { ViteHubCliCommandNamespace, ViteHubCliContext, ViteHubCliSpawn, ViteHubCliSpawnOptions } from "@vite-hub/internal/cli"
+import type { ViteHubCliCommandNamespace, ViteHubCliContext } from "@vite-hub/internal/cli"
+
+interface ViteHubCliSpawnResult {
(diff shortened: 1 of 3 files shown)


onmax:
**Head:** `e9d2bb562210619d1adddbfdbe4484d26d7a2630`
**Verdict:** proceed

**Direction:** Keep namespace collection and the canonical contributor contract in `@vite-hub/internal/cli`, while retaining thin self-contained public type adapters where published declaration compatibility requires them.

**Evidence:** The CLI now delegates collection to the tested internal collector. Exact-head review showed that fully replacing public structural types would tighten Database's callback context and could expose a private package in CLI declarations, so the repair preserves those published shapes without duplicating collection behavior.

**ViteHub precedent:** Published packages bundle `@vite-hub/internal` via `alwaysBundle`, but public package declarations remain independently consumable.

**Strongest challenge:** Total type deduplication conflicts with the stated no-public-API-change requirement. The minimal compatible boundary keeps only declaration adapters public and verifies their implementations against the shared contributor contract with `satisfies`.

**Next action:** Resolve the addressed exact-head threads, run the new CI head, then obtain a fresh exact-head Codex review before merging.

onmax:
@codex review

onmax:
@codex review

PR #739, opened 2026-07-20
feat: complete structured error contracts

onmax:
Consolidates the surviving Agent, Workflow, and Queue structured-error work onto current `main` after the original stacked pull requests lost their base branches. Built-in errors now use closed codes, code-derived messages, allowlisted details, private causes, immutable serialization, and provider-aware operation and retry handling; application-defined Workflow errors retain explicit JSON-safe messages and details.

This replaces #712, #713, and #715 and folds in their closed prerequisites #694, #696, and #706 without carrying the obsolete stack history. It also preserves the current OpenWorkflow acquisition, reset, cleanup, and worker lifecycle behavior inside the closed contract.

PR #744, opened 2026-07-20
fix: harden runtime and verification edges

onmax:
Hardens four existing runtime and verification boundaries. Chat senders are now authorized from minimal identity metadata before attachment URLs are fetched, transcription downloads cancel and release oversized or erroring streams, and Cloudflare Queue rejects unsupported `region` options for single sends and batches.

The docs typecheck now runs `vue-tsc` through an exact Docus 5.12.3 diagnostic allowlist, so compiler crashes, changed dependency diagnostics, and application type errors remain failures instead of disappearing through a grep pipeline. The Queue change is integrated with the structured error contract from #739, including typed and normalized unsupported-option details.

### DIFF
--- a/docs/content/docs/server-primitives/queue.md
+++ b/docs/content/docs/server-primitives/queue.md
@@ -387,7 +387,7 @@ When migrating from the first structured Queue error contract, add `custom: true
 | `QUEUE_PROVIDER_RESPONSE_INVALID` | A successful Vercel send returned a missing or malformed `messageId`. |
 | `CLOUDFLARE_BINDING_RESOLUTION_REQUIRED` | A direct Cloudflare client was created without a concrete binding. |
 | `CLOUDFLARE_BINDING_INVALID` | The Cloudflare binding does not expose `send()` and `sendBatch()`. |
-| `CLOUDFLARE_UNSUPPORTED_ENQUEUE_OPTIONS` | Cloudflare received unsupported enqueue options such as `idempotencyKey` or `retentionSeconds`. |
+| `CLOUDFLARE_UNSUPPORTED_ENQUEUE_OPTIONS` | Cloudflare received unsupported enqueue options: `idempotencyKey`, `region`, or `retentionSeconds`. |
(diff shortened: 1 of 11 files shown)


onmax:
**Head:** `9319bc8eb52d342c21bed4a6176eefaaddf3af5a`

**Verdict:** proceed

**Direction:** Keep the repair scoped to the four boundaries named by the owner: authorize minimal chat identity before remote attachment hydration while preserving documented message and channel/chat context for custom invoker resolution; cancel and unlock transcription response streams on size/read failures; reject Cloudflare unsupported `region` enqueue option consistently for single and batch sends; and replace the docs grep-based typecheck suppression with a fail-closed, exact diagnostic allowlist.

**Evidence:** The exact-head diff uses a metadata/text-only UI message to resolve authorization actors before attachment hydration, adds regression assertions for invoker message and channel/chat context, adds cancellation/release tests for response streams, extends the structured Cloudflare unsupported-option contract and runtime tests, makes docs typechecking fail on crashes, changed diagnostics, and non-Docus errors, and explicitly documents all three Cloudflare-unsupported enqueue options. The current base is an ancestor of the head and the PR is cleanly mergeable. Full CI passed on the preceding head; checks are running for the Codex-finding repair.

**ViteHub precedent:** Existing Agent access capabilities centralize chat admission in `isChatMessageAuthorized`; `createChatMessageTriggerInput` defines the documented invoker input shape; queue provider adapters own provider support; #739 established typed normalized queue errors; repository CI uses package-owned scripts for verification boundaries.

**Strongest challenge:** Early authorization must not fetch attacker-controlled attachment URLs, but custom invoker resolution still needs the current message and channel/chat context. The authorization path now supplies that shape from webhook metadata and text only, then hydrates attachments only after admission.

**Next action:** Await exact-head checks and one new exact-head Codex review, address any terminal findings, and merge only when all gates are satisfied.

onmax:
@codex review

onmax:
@codex review

PR #636, opened 2026-07-16
feat(agent)!: rename chatTitle to title

onmax:
Renames `chatTitle()` to `title()` and generalizes its public types, capability id, stream data parts, documentation, and Channel delivery internals. The Capability can now describe its full boundary: applications can use the finish extension to name durable jobs or runs, while compatible message Channels can still deliver the same title to a thread.

This is intentionally breaking while the Agent API is in active development. Consumers now import `title`, read the `title` extension, and handle `data-title` / `type: "title"` stream metadata. The fallback becomes `Untitled`, with existing model, heuristic, harness-driver, stream, and once-per-thread delivery coverage preserved.

PR #748, opened 2026-07-20
fix(agent): mark failed Channel titles

onmax:
Marks failed Agent Invocation titles through the existing Channel title effect. The normal title is still generated and applied first; terminal failures then reuse that cached title for one idempotent, max-length-safe `ERROR: ` rename, so Channel conversations expose the failure without another model call. If generation fails, the title falls back to `ERROR: Untitled`, while `trigger` and `when` opt-outs remain respected.

Keeps the successful path and generic Channel boundary unchanged, with Discord retaining its existing 100-character platform cap. Coverage exercises ordering, failed early-delivery retries, nested prefixes, configured length limits, fallback behavior, title filters, and the Discord adapter mutation.

The workspace quota failure remains separate. Its follow-up seam is trusted-host root allocation and file-by-file materialization in `packages/workspace/src/session/trusted-host.ts`, coordinated through `packages/workspace/src/session/harness.ts`; because #746 replaces that session model, the storage follow-up should reconcile with that PR instead of landing a speculative redesign here.

### DIFF
--- a/packages/agent/src/capabilities/title.ts
+++ b/packages/agent/src/capabilities/title.ts
@@ -7,6 +7,7 @@ import {
   finishMessageChannelTitleDelivery,
   messageChannelStateContextKey,
   messageChannelTitleDeliveredContextKey,
+  resetMessageChannelTitleDelivery,
 } from "../internal/channels.ts"
 import { getMessageText } from "../messages.ts"
 import { normalizeAgentDriver } from "../internal/agent-driver.ts"
@@ -573,6 +574,7 @@ export function title<TRuntimeConfig extends AgentRuntimeConfig = AgentRuntimeCo
         await finishMessageChannelTitleDelivery(await getChannelDeliveryAttempt(), false).catch(() => undefined)
       }
       let title: Promise<string | undefined> | undefined
+      let titleClaimed = false
       let titleSkipped = false
       const getTitle = () => {
         title ??= (async () => {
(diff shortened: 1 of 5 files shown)


onmax:
**Head:** `c35a120095efb28ac35415af93a8fdcfc7a4d56e`
**Verdict:** proceed

**Direction:** Reuse the existing Agent Definition `title` capability and Channel finish-effect pipeline to apply the ordinary generated title first, then an idempotent, length-bounded `ERROR: ` title on terminal failure. Preserve title trigger/`when` opt-outs, claim ownership, successful retry behavior, and adapter-specific limits.

**Evidence:** The PR title/body request failed Channel titles. Exact-head review exposed two state-backed ownership defects; this head now suppresses failure titles when the invocation did not own delivery and clears/non-persists the delivery marker for failed runs. The focused state primitive regression verifies marker cleanup.

**ViteHub precedent:** #636 established `title` as the Channel-scoped capability; #736 uses the same claim, delivery-attempt, and finish-effect primitives for once-per-thread Channel titles.

**Strongest challenge:** A failed run must display its error title without permanently consuming the once-per-thread marker or racing an invocation that owns the title claim. The failure path now requires ownership, keeps the claim through ordered effects, and leaves durable delivery unmarked so a successful retry can replace the error title.

**Next action:** Await exact-head CI, resolve the addressed review threads, and request one exact-head Codex review.

onmax:
@codex review

onmax:
@codex review

PR #751, opened 2026-07-20
fix(agent): preserve channel attachment references by default

onmax:
Preserves Channel-hosted image, audio, and generic file attachments as typed `AttachmentPart` references by default. Shared message normalization now carries every provider handle together—URL, lazy `fetchData`, stable `fetchMetadata`, media type, name, and size—while bounded text attachments keep their existing decoded-text behavior. This removes the capability opt-in and avoids duplicate audio-plus-prose fallbacks.

The AI SDK converter passes inline data and validated HTTPS references directly to supported model inputs, and model-backed Agent Drivers resolve provider callbacks only when the model invocation needs bytes. Resolution is bounded to 25 MiB by default through `execution.attachments.maxBytes`; synchronous conversion fails explicitly for callback- or Blob-only inputs, while harness consumers omit callback/Blob-only references and retain safe URLs without implicit server fetches or temporary-file materialization. The Channel docs describe expiry, authentication, serialization, model-disclosure, protocol, and byte boundaries.

Validation: exact-head CI passes; 425 focused provider/runtime tests and 274 focused runtime tests pass locally; `pnpm --filter @vite-hub/agent typecheck` passes; and the agent package build passes with publint clean. Exact-head Codex review found no major issues after Web Chat metadata classification, user-role/current-turn callback resolution, safe media-type handling, Blob handling, and harness callback boundaries were corrected.

### DIFF
--- a/docs/content/docs/agents/agent-definitions.md
+++ b/docs/content/docs/agents/agent-definitions.md
@@ -113,7 +113,7 @@ export default defineAgent({
 })
 ```
 
-The callback also receives the invocation input and runtime handles. Capabilities that contribute Agent Triggers, chat admission or attachments, or static Workspace Sources must stay in a static array because ViteHub registers those contributions before an invocation starts.
+The callback also receives the invocation input and runtime handles. Capabilities that contribute Agent Triggers, chat admission, or static Workspace Sources must stay in a static array because ViteHub registers those contributions before an invocation starts.
 
 ## Add Workspace context
 
--- a/docs/content/docs/agents/channels.md
+++ b/docs/content/docs/agents/channels.md
@@ -147,7 +147,7 @@ export default defineAgent({
 })
 ```
 
-Model-backed Agents pass inline data and HTTPS references as typed model input. When a part has `fetchData`, the model driver prefers that adapter-owned callback immediately before invocation and checks both the declared and resolved byte size. The default limit is 25 MiB; change the real resource policy with `driver.execution.attachments.maxBytes`.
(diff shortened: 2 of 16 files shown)


onmax:
Verdict: proceed
Head: `a10fc82c6579039b4e421f65a2588303407bff8d`
Direction: Preserve Channel-hosted non-text attachments as typed Agent Message references during shared normalization, then let the consuming Agent Driver or Capability choose whether to forward a validated HTTPS URL or resolve provider-owned bytes within an explicit limit.
Evidence: Chat SDK adapters can supply `data`, `fetchData`, `fetchMetadata`, `url`, media type, name, and size together; the prior normalization kept only audio and selected exactly one handle, so image/file references and fallback handles were lost before any consumer could choose safely.
Philosophy: Repository guidance says to preserve core Agent Definition/Capability/Channel boundaries, keep authority explicit, and make runtime state inspectable. The generated ViteHub skill routes Channel message normalization through Channels and Triggers while requiring authority to remain visible in Capabilities, Workspaces, Sources, or Channels.
ViteHub precedent: `AudioPart` already carried provider callbacks and metadata; `transcribe()` already resolves bytes at consumption time with `maxBytes`; harness serialization already treats callbacks as a boundary rather than silently materializing them.
Challenge: A default could appear to grant models or harnesses attachment access or create SSRF/persistence behavior. It does neither: normalization is inert, only HTTPS URLs are forwarded, provider callbacks are resolved only for the latest Channel user turn at model invocation, harnesses omit callback/Blob-only references, and byte limits are enforced before and after resolution. Unknown or incompatible media types remain generic file references rather than receiving invented typed metadata.
Next: Wait for exact-head CI and one exact-head Codex review, address any terminal findings, then merge.

onmax:
@codex review

onmax:
@codex review

onmax:
@codex review

onmax:
@codex review

onmax:
@codex review

onmax:
@codex review

onmax:
@codex review

PR #752, opened 2026-07-20
fix(blob): make private Vercel storage runtime-complete

onmax:
Private Vercel Blob output selected the lazy optional-peer driver and externalized its implementation, so isolated functions omitted the runtime dependency even when `@vite-hub/blob` selected the store.

This makes `@vercel/blob` a Blob-owned runtime dependency, selects a direct Vercel driver only for Vercel output, and copies the SDK into the isolated function. `files-sdk` remains an optional peer for generic adapters, avoiding its incompatible optional `h3@^1` peer in normal Blob installs. The regression packs a clean consumer that installs only Blob and Vite with strict peer enforcement, builds the generated Vercel function, and executes `GET /api/stats` against a local Blob API while the existing optional-peer coverage preserves generic non-Vercel reachability.

### DIFF
--- a/packages/blob/fixtures/private-vercel-consumer/src/server.ts
+++ b/packages/blob/fixtures/private-vercel-consumer/src/server.ts
@@ -1,0 +1,3 @@
+import { blob } from "@vite-hub/blob"
+
+export default async () => Response.json(await blob.list())
--- a/packages/blob/fixtures/private-vercel-consumer/vite.config.ts
+++ b/packages/blob/fixtures/private-vercel-consumer/vite.config.ts
@@ -1,0 +1,16 @@
+import { hubBlob } from "@vite-hub/blob/vite"
+import { defineConfig } from "vite"
+
+export default defineConfig({
+  blob: {
+    access: "private",
+    driver: "vercel-blob",
+    token: "vercel_blob_rw_test",
+  },
+  build: {
+    outDir: "dist/client",
+    rollupOptions: { input: "src/server.ts" },
+    ssr: "src/server.ts",
+  },
+  plugins: [hubBlob()],
+})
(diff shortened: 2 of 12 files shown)


onmax:
**Head:** `5c9d931648bd845ccc89d1428610eec67b8f0001`
**Verdict:** proceed

**Direction:** Make private Vercel Blob runtime-complete inside `@vite-hub/blob`: own `@vercel/blob`, use the direct Vercel driver only for Vercel output, and copy the SDK into the isolated function while retaining `files-sdk` as an optional peer for generic adapters.

**Evidence:** The exact-head diff is a single Vercel-only commit atop current `main`. Blob build/publint, focused missing-object/store-error regression, typecheck, and the preceding head's complete CI passed through every finished gate. Codex's store-error finding is addressed by matching `BlobNotFoundError` and object-specific messages without swallowing missing-store failures.

**ViteHub precedent:** Provider-output generation already copies host-required SDKs into isolated Vercel functions. This applies that ownership boundary to the direct Vercel Blob SDK without exposing generic `files-sdk` adapter plumbing.

**Strongest challenge:** The direct provider driver duplicates mapping behavior from the generic Files adapter and must preserve every Blob contract. It remains justified because making `files-sdk` mandatory introduces its incompatible optional `h3@^1` peer into normal Blob installs.

**Next action:** Obtain exact-head CI and Codex results, address any actionable feedback, then merge.

onmax:
@codex review

onmax:
@codex review

onmax:
@codex review

onmax:
@codex review

onmax:
@codex review

onmax:
@codex review

onmax:
@codex review

onmax:
@codex review

PR #747, opened 2026-07-20
feat: add built-in deployment presets

onmax:
Adds a required built-in deployment `preset` to the root `vite-hub` facade and resolves Cloudflare, Netlify, Vercel, Deno Deploy, and Node/VPS through one typed deployment plan. That plan now drives Nitro selection, service adapters and guarantees, output validation, and a generated deployment manifest; unsupported capabilities and conflicting provider selection fail during configuration instead of silently degrading.

Moves Deno native dependency staging and non-interactive create-or-update deployment into the generated output, including Sharp optional native packages, `deno.json`, and a package-owned deploy runner. The runner uploads a complete temporary copy outside the consumer repository so ignored `.output` files remain deployable, keeps bundled Sharp native packages resolvable from the output root, and applies native-module upload semantics to both deployment paths. Root feature options are reduced to plan-driven booleans where appropriate, fixtures and docs use the required preset, and real-plugin plus Drop matrix coverage switches the same application across all five targets.

PR #753, opened 2026-07-20
fix(deploy): upload native modules on Deno updates

onmax:
Stacks on #747’s generated Deno runner. Keeps `--allow-node-modules` in the shared deploy arguments so an ordinary existing-app update uploads the staged `node_modules` tree just like the create path; without it, the external upload root fixes ignored `.output` files but silently drops Sharp’s native packages on subsequent deployments.

The generated-runner regression exercises both paths from a consumer whose `.gitignore` excludes `.output`, verifies the complete temporary stage includes the Linux x64 Sharp and libvips artifacts, and requires native-module upload semantics on update and create.

### DIFF
--- a/packages/internal/src/build/deno-runtime-packages.ts
+++ b/packages/internal/src/build/deno-runtime-packages.ts
@@ -291,10 +291,10 @@ try {
   }
   await cp(sourceRoot, uploadRoot, { recursive: true })
 
-  const common = ["--org", organization, "--app", app]
+  const common = ["--allow-node-modules", "--org", organization, "--app", app]
   const deployment = await run(["deploy", ".", "--prod", ...common], uploadRoot)
   if (deployment.code !== 0) {
-    const creation = await run(["deploy", "create", ".", "--source", "local", "--do-not-use-detected-build-config", "--runtime-mode", "dynamic", "--allow-node-modules", "--entrypoint", "server/index.ts", "--working-directory", ".", "--region", region, ...common], uploadRoot)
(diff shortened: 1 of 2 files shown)


onmax:
**Head:** `8d50d36332851a92600f3608f094db057e3289a7`
**Verdict:** proceed
**Direction:** Keep the repair scoped to sharing Deno Deploy's existing `--allow-node-modules` argument across update and create, with the generated-runner regression covering both paths.
**Evidence:** The title/body identify native packages being omitted only on ordinary existing-app updates. The exact diff moves the existing flag into the shared argument list and updates the focused end-to-end runner assertion.
**ViteHub precedent:** Base commit `68e7bc6` introduced native-module preservation and the current base's Deno runner already stages ignored output plus Sharp/libvips artifacts; shared deploy arguments are the obvious boundary for behavior required by both invocation paths.
**Strongest challenge:** The ordinary `deploy` subcommand must accept `--allow-node-modules`; the generated runner now invokes that exact argument path and requires it in the regression.
**Next action:** Run focused verification and review the exact diff, then request exact-head Codex review.

onmax:
@codex review

onmax:
Closing as obsolete after exact-head review. This exact shared-flag change previously landed in 68e7bc6 and was corrected by 790ab5a because `--allow-node-modules` is only accepted by `deno deploy create`; passing it to the existing-app `deno deploy` path fails argument parsing and then incorrectly falls back to creating an app that already exists. The current Deno CLI exposes no supported existing-app equivalent, so this branch cannot fulfill the stated intent without reintroducing that known regression.

PR #731, opened 2026-07-19
refactor: simplify runtime implementation

onmax:
Removes compatibility and duplicate implementation paths across Agent, Schedule, Workflow, and Workspace. Agent Definitions are now the only Agent runtime input, the AI SDK adapter has one execution option, synchronous DevTools metadata preserves authored instructions without maintaining a second renderer, and Workspace source wrappers no longer delete properties their factories never produce.

Moves default-export and object-property scanning into the existing internal source scanner, so Schedule discovery and provider output share one implementation and Workflow reuses the same literal masking. Discord Gateway generation now uses the common Agent deployment catalog. The resulting production and fixture changes remove roughly 730 net lines while preserving the current package contracts covered by the affected suites.

PR #755, opened 2026-07-20
fix(internal): stop source scanner recursion

onmax:
Prevents the source scanner from recursively reclassifying control-flow regexes while it rescans template expressions and nested matching regions. A classification cache is scoped to each active scan, shared through nested scans, preserves completed results, and prevents repeated candidates from causing exponential work.

Structural-regex regressions reproduce the Workflow closeBundle stack overflow, repeated-rescan hang, and nested-template recursion while verifying that delimiters inside regex literals remain non-structural. The focused scanner suite passes 21 tests, internal typecheck and affected-file lint pass, and hosted CI verifies the full repository.

### DIFF
--- a/packages/internal/src/source-scanner.ts
+++ b/packages/internal/src/source-scanner.ts
@@ -160,7 +160,7 @@ function isControlFlowRegexStart(source: string, index: number, controlFlowRegex
 
     for (let current = closeParen; current >= 0; current--) {
       if (source[current] !== "(") continue
-      if (findMatching(source, current, "(", ")") !== closeParen) continue
+      if (findMatching(source, current, "(", ")", controlFlowRegexes) !== closeParen) continue
       const head = source.slice(0, current).replace(/\/\*[\s\S]*?\*\/|\/\/[^\n]*/g, " ")
       const result = /(?:^|[^\w$])(?:catch|for|if|while|with)\s*$/.test(head)
       controlFlowRegexes.set(index, result)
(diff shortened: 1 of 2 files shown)


onmax:
**Head:** `4365946cbd2db59c3f9b23df40f5a466e950ddb9`
**Verdict:** proceed

**Direction:** Keep the repair scoped to preventing recursive reclassification of control-flow-regex slashes while sharing cached classifications through every nested matching scan so repeated and nested candidates stay bounded.

**Evidence:** The exact-head diff changes only `packages/internal/src/source-scanner.ts` and its focused tests. In-progress cache entries terminate same-candidate cycles, completed boolean entries prevent exponential repeated rescans, and `findMatching` now shares that cache through nested template conditions. The focused 21-test scanner suite, internal typecheck, affected-file lint, and prior hosted CI pass.

**ViteHub precedent:** The scanner was introduced and then simplified in `b3f7238` (`refactor: simplify runtime implementation (#731)`); this repair extends that existing internal primitive instead of introducing a parser or new abstraction.

**Strongest challenge:** Cached classifications must remain local enough not to reuse offsets across different sources or turn division into regex. Each public scan path creates its own cache, all recursive scans over that source share it, and division, repeated-candidate, and nested-template regressions cover those boundaries.

**Next action:** Wait for exact-head checks and obtain one clean Codex review for this head.

onmax:
@codex review

onmax:
@codex review

onmax:
@codex review

PR #761, opened 2026-07-20
fix(queue): bound Cloudflare resource names

onmax:
Cloudflare Queue names can exceed the provider's 63-character limit when a full deployment scope is combined with the encoded logical name. This keeps every valid existing encoded name unchanged, reuses the canonical readable deployment-logical identity when safe, and otherwise derives a bounded readable SHA-256 name inside the Queue provider boundary.

The exact Drop topology now builds with its existing Worker, R2 bucket, Sandbox container, Durable Object, and legacy Queue identities unchanged. Focused collision and determinism contracts, the packed Cloudflare consumer, strict Worker reachability scans, and Vercel, Deno, Netlify, and Node config assertions are green.

Stacked on #747; review after that provider-preset branch.


onmax:
**Head:** `9394266c9788ec8c964205c2a69a53cffcfae4d7`
**Verdict:** proceed

**Direction:** Bound physical Cloudflare Queue names inside the Queue provider boundary, retain existing encoded identities when they fit, and preserve the canonical readable deployment/logical identity only when it is safe and unambiguous.

**Evidence:** The base implementation concatenates the full prefix with a hex-encoded logical name and throws above Cloudflare's 63-character limit. The deployment preset compensates by truncating the prefix, losing deployment identity. This head moves bounded derivation into the Cloudflare Queue naming primitive, restores the full deployment prefix, detects derived-name collisions, and carries the physical-to-logical map into both generated and playground Cloudflare consumers. Direct workers still decode encoded and readable names but refuse to guess digest identities when the map is omitted.

**ViteHub precedent:** Provider-specific naming belongs behind ViteHub's provider integration boundary; deployment presets should pass complete ViteHub deployment identity instead of encoding provider limits.

**Strongest challenge:** Reusing a readable legacy Queue could collide with the digest namespace or select an unintended pre-existing resource. The implementation limits reuse to the established lowercase `deployment-logical` shape, excludes digest-shaped names, checks the provider limit, separates hash inputs, rejects derived collisions, and uses explicit maps for digest dispatch.

**Next action:** Wait for exact-head CI, resolve the addressed Codex thread, and request a new exact-head Codex review.

onmax:
@codex review

onmax:
@codex review

onmax:
@codex review

PR #764, opened 2026-07-20
fix(blob): propagate Vercel runtime to Nitro

onmax:
Propagates the detected Vercel hosting provider into the Nitro shared Blob runtime so private Vercel Blob selects the statically bundled driver in the function that imports it. This keeps the generic optional-peer driver for plain Vite and non-Vercel runtimes while preserving the existing isolated provider output.

Refs #763

PR #759, opened 2026-07-20
fix(blob): keep Cloudflare R2 provider-native

onmax:
Cloudflare Nitro builds still routed Blob through the mixed R2 fallback graph and Vercel package staging. Clean consumers therefore tried to copy absent AWS SDK peers after Nitro finished, while the Worker retained `files-sdk` and `files-sdk/r2` reachability it could never use with a native R2 binding.

This selects a dedicated binding-native R2 driver only for Cloudflare output, keeps Blob signing runtime-complete through the Blob-owned `aws4fetch` dependency, and skips and cleans Blob-owned Vercel function output when Nitro owns Cloudflare deployment output. Generic, bindingless, and cross-provider R2 use keeps the existing lazy HTTP fallback and optional peers.

The regression packs a production-only consumer with no direct Files SDK or AWS SDK dependencies, builds the generated Cloudflare runtime, executes native signing, and asserts that the output contains neither `files-sdk` nor AWS SDK reachability. The exact Drop `c42ec504` + Queue `8fcbc57` Cloudflare build also completes with no Blob Files/AWS reachability, no `.vercel` output, and unchanged Queue prefix, logical mapping, and R2 bucket identity.

Validated with `@vite-hub/blob` build/publint, typecheck, and all 133 package tests. This PR is stacked on #752 so its diff contains only the Cloudflare R2 provider-output correction.

PR #767, opened 2026-07-20
fix(blob): restore hosted provider output contracts

onmax:
Restores the hosted Blob provider behavior that was proven in #759 and #764 but did not reach `main` when their stacked branches closed after #747 merged.

Private Vercel Blob now propagates the hosting provider into Nitro's shared runtime and selects the statically bundled `@vercel/blob` driver. Cloudflare Nitro output selects a binding-native R2 driver, owns signing through `aws4fetch`, and excludes Files SDK and AWS SDK fallback reachability while generic and bindingless R2 use keeps the existing lazy fallback.

Provider-output cleanup is serialized and ownership-marked. Blob removes only unchanged isolated functions it created, preserves shared Nitro `__server.func` output, and re-evaluates ownership inside the shared output queue so concurrent writers cannot be deleted from a stale marker read.

Shared sibling Vercel output stages Blob provider dependencies before the serialized function writer, so plugin order cannot drop runtime packages. The packed consumer regression builds and executes private Vercel Blob without consumer provider dependencies, then builds Cloudflare R2 in the same root and verifies native reachability plus preservation of the prior shared Vercel function. The exact Drop `c42ec504` composition with Queue `4ad1e19` passes fresh-store install, typecheck, standalone Vercel `/api/stats`, subsequent Cloudflare build, resource identity checks, and negative Files SDK, AWS SDK, MCP, and PKCE reachability. Blob passes 141 tests with no type errors; Internal passes 157 tests; repository lint, full typecheck, and 24 contract tests pass.

### DIFF
--- a/packages/blob/fixtures/cloudflare-r2-consumer/vite.config.ts
+++ b/packages/blob/fixtures/cloudflare-r2-consumer/vite.config.ts
@@ -1,0 +1,18 @@
+import { hubBlob } from "@vite-hub/blob/vite"
+import { defineConfig } from "vite"
+
+export default defineConfig({
+  build: {
+    outDir: "dist/cloudflare",
+    rollupOptions: {
+      external: ["cloudflare:workers"],
+      input: ".vitehub/nitro/blob/runtime.mjs",
+      output: { entryFileNames: "server.mjs" },
+    },
+  },
+  nitro: { preset: "cloudflare_module" },
+  plugins: [
+    { name: "nitro:main" },
+    hubBlob({ binding: "BLOB", bucketName: "assets", driver: "cloudflare-r2" }),
+  ],
+})
--- a/packages/blob/package.json
+++ b/packages/blob/package.json
@@ -34,6 +34,7 @@
     "./drivers/azure": "./dist/drivers/azure.js",
     "./drivers/box": "./dist/drivers/box.js",
     "./drivers/cloudflare": "./dist/drivers/cloudflare.js",
+    "./drivers/cloudflare-native": "./dist/drivers/cloudflare-native.js",
     "./drivers/digitalocean-spaces": "./dist/drivers/digitalocean-spaces.js",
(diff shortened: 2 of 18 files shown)


onmax:
**Head:** `6634cb40c52ea08d35fcee902922c38a3da85758`

**Verdict:** `proceed`

**Direction:** Restore the hosted Blob provider output contracts lost when the stacked Blob branches were closed: private Vercel Blob must select its bundled provider driver in Nitro, Cloudflare output must use binding-native R2 without optional fallback reachability, and Blob cleanup must preserve shared provider output.

**Evidence:** The PR remains based directly on `main`; its diff is confined to Blob provider/runtime/build behavior, the shared deployment-output queue, fixtures, tests, and dependency metadata. PR #759 proves the Cloudflare-native R2 contract and PR #764 proves the Vercel runtime propagation contract. Exact-head Codex review on `c57d66b` identified that a sibling Vercel output could consume the advertised Blob runtime during a Cloudflare Nitro build without receiving its optional R2 dependencies. The current head stages those dependencies before the sibling writer runs, relying on the shared writer preservation of `node_modules`, and adds an ordering regression.

**ViteHub precedent:** Existing Blob code keeps generic and bindingless R2 on lazy optional fallbacks while provider builds specialize deployment output; `@vitehub/internal` owns host-independent serialized deployment-output mutation. Composed provider runtimes already use the shared product registry to coordinate sibling output.

**Strongest challenge:** The repair combines Vercel propagation, Cloudflare specialization, and output ownership, and cross-provider composition can create output that Blob does not itself emit. The shared runtime registry is therefore the authoritative signal for dependency staging; the new probe verifies that contract without adding another ownership mechanism.

**Next action:** Wait for checks on `6634cb4`, resolve the addressed Codex thread, then obtain a new exact-head Codex review before merge.

onmax:
@codex review

onmax:
@codex review

onmax:
@codex review

PR #769, opened 2026-07-20
fix(blob): authenticate Vercel object reads

onmax:
Reads Vercel Blob objects through the authenticated SDK using the canonical URL returned by metadata lookup. This prevents private objects from being fetched anonymously when a per-write access override differs from the normalized store default, while keeping the application and provider configuration unchanged.

Adds bidirectional coverage for public/private access mismatches and explicitly guards against the anonymous fetch path that produced the hosted 403. The stateless access recovery costs one metadata request, which is the provider-authoritative way to select the immutable Vercel store access mode.

### DIFF
--- a/packages/blob/src/drivers/vercel-bundled.ts
+++ b/packages/blob/src/drivers/vercel-bundled.ts
@@ -29,6 +29,16 @@ function isMissingBlobError(error: unknown) {
   )
 }
 
+function resolveBlobAccess(url: string, fallback: "private" | "public") {
+  try {
+    const hostname = new URL(url).hostname
+    if (hostname.endsWith(".private.blob.vercel-storage.com")) return "private"
+    if (hostname.endsWith(".public.blob.vercel-storage.com")) return "public"
+  }
+  catch {}
+  return fallback
+}
+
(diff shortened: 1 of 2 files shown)


onmax:
**Head:** `1e688dc7eab21b63be4ecdd9ca19ac01a0640c88`
**Verdict:** proceed

**Direction:** Read every Vercel Blob object through the authenticated `@vercel/blob` SDK, using provider metadata to recover the object access when a per-write override differs from the store default.

**Evidence:** The exact-head diff is one focused commit atop current `main`, limited to the bundled Vercel driver and a bidirectional regression. The driver resolves metadata by pathname, passes its canonical URL and authenticated token to `get`, preserves timeout and missing-object behavior, and explicitly avoids anonymous `fetch`.

**ViteHub precedent:** PR #752 made the bundled `@vercel/blob` driver the Vercel-output runtime boundary and already uses provider metadata and authenticated SDK operations to preserve Blob contracts. This repair stays inside that provider-owned primitive.

**Strongest challenge:** Access recovery relies on Vercel canonical `.private.blob.vercel-storage.com` and `.public.blob.vercel-storage.com` hostnames. The inference is narrowly suffix-matched and falls back to the normalized store access for unknown or malformed URLs, while metadata remains the provider-authoritative lookup.

**Next action:** Run focused verification and the base-branch code review, fix actionable findings, then obtain exact-head CI and Codex review before merging.

onmax:
@codex review

PR #771, opened 2026-07-20
fix(deploy): upload native modules on Deno updates

onmax:
Existing Deno app deployments used the ordinary update command without `--allow-node-modules`, so ViteHub's staged native dependencies were omitted from uploads and Sharp failed at runtime. Apply the upload flag to both create and update, and remove the manual `DENO_DEPLOY_NODE_MODULES_ENABLED` assertion that exposed provider policy to users.\n\nCover the generated runner directly and through a real Nitro/ViteHub Deno consumer. The consumer simulates an existing app and uploader filtering, copies the generated artifact, then executes it with Deno and performs a real Sharp PNG transform.

### DIFF
--- a/packages/internal/src/build/deno-runtime-packages.ts
+++ b/packages/internal/src/build/deno-runtime-packages.ts
@@ -331,12 +331,9 @@ try {
   }
   await cp(sourceRoot, uploadRoot, { recursive: true })
 
-  const common = ["--org", organization, "--app", app]
-  const creation = await run(["deploy", "create", ".", "--source", "local", "--do-not-use-detected-build-config", "--runtime-mode", "dynamic", "--allow-node-modules", "--entrypoint", "server/index.ts", "--working-directory", ".", "--region", region, ...common], uploadRoot)
+  const common = ["--allow-node-modules", "--org", organization, "--app", app]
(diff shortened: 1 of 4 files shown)


onmax:
**Head:** `3b07b37935cfdf0f1088150a5817b32191742966`
**Verdict:** proceed
**Direction:** Ensure Deno Deploy uploads ViteHub's staged native Node modules on both app creation and existing-app updates, without requiring users to attest to provider policy through a ViteHub-specific environment variable.
**Evidence:** The generated runner already stages runtime packages and optional native dependencies; creation passes `--allow-node-modules`, while the fallback update command on the base branch omits it. The PR applies the same Deno CLI upload flag to both paths and tests a real Sharp transform from a copied deployment artifact.
**ViteHub precedent:** Deployment presets own host-specific command construction behind ViteHub's generated runner, while public configuration stays in ViteHub language and generated artifacts remain directly inspectable.
**Strongest challenge:** The consumer test hard-codes Sharp's Linux x64 native artifacts, so it must not execute on platforms where those optional packages are unavailable.
**Next action:** Wait for checks on the platform-guard repair, resolve the addressed review thread, and obtain an exact-head Codex review.

onmax:
@codex review

onmax:
@codex review

PR #770, opened 2026-07-20
fix(queue): close blob runtimes in provider outputs

onmax:
Closes provider-selected Blob runtimes inside Queue-owned provider bundles. Queue now participates in the shared composed-output lifecycle and aliases the selected Blob runtime into both the root Queue function and per-definition callback functions for Vercel and Cloudflare, so Queue Definitions can use Blob without adding provider dependencies to the application.\n\nBlob's files-sdk driver registry now exhaustively records each driver's runtime packages. The public vite-hub facade owns those packages, while generated Vercel functions stage only the packages selected by configured stores and fail the build when a selected runtime is unavailable.\n\nThe regression runs a real Vite lifecycle with hubBlob() and hubQueue(), checks both isolated Vercel outputs for closed runtime reachability, executes a GCS-backed copied callback from an OS temporary directory, and proves an inactive Azure runtime is absent. The strict packed vite-hub consumer resolves the selected GCS runtime with only vite-hub declared. Exact Drop packed builds also pass for Vercel and Cloudflare; Cloudflare remains free of Vercel, AWS SDK, MCP, and PKCE reachability.

PR #773, opened 2026-07-21
fix(queue): preserve Nitro Cloudflare bindings

onmax:
Restores dynamically discovered Queue bindings in Nitro-owned Cloudflare output. Queue no longer runs its entire Vite Integration as a post plugin, so `nitro:init` snapshots the generated bindings while Queue provider-output closure still observes the shared catalog populated during `buildEnd`.

Expands the real Vite and Nitro consumer regression to prove the exact long Drop deployment Queue identity reaches final Wrangler output alongside manual Queue, Rate Limit, R2, container, and Worker-name configuration. The emitted server contains the Cloudflare Queue hook while Vercel Queue and Functions imports and Vercel output remain absent.

### DIFF
--- a/packages/queue/src/vite.ts
+++ b/packages/queue/src/vite.ts
@@ -117,7 +117,6 @@ export function hubQueue(options?: QueueModuleOptions): QueueVitePlugin {
   let validatesNitroDefinitions = false
 
   return {
-    enforce: "post",
     name: "@vite-hub/queue/vite",
     vitehub: {
       cli: async () => {
--- a/test/consumer/deployment-presets.test.ts
+++ b/test/consumer/deployment-presets.test.ts
@@ -7,14 +7,8 @@ import { delimiter, dirname, join, resolve } from "node:path"
 import { promisify } from "node:util"
 
 import { build, copyPublicAssets, createNitro, prepare, prerender } from "nitropack/core"
-import {
-  build as buildNitro3,
-  copyPublicAssets as copyNitro3PublicAssets,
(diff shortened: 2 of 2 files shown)


onmax:
Head: `c8b4f403763f7224bfdafa17813fa23b56d11ec2`
Verdict: proceed
Direction: Restore dynamically discovered Queue bindings for developers building Nitro-owned Cloudflare Workers by letting Queue participate in normal Vite integration ordering while retaining the shared provider-output closure at the Queue integration boundary.
Evidence: PR #770 added whole-plugin `enforce: "post"` together with the provider-output closure. In a real Vite + Nitro build, Nitro's initialization snapshots Wrangler configuration before the post-ordered Queue integration has populated generated Queue bindings; removing only that ordering override preserves the closure and lets the generated consumer and producer reach final `.output/server/wrangler.json`.
ViteHub precedent: Queue, Rate Limit, and Blob contribute host-independent provider output through `useComposedProviderOutput` and compose it at the integration that owns the runtime; existing integrations do not require whole-plugin post ordering to preserve their contributions.
Strongest challenge: The post ordering might be required for the PR #770 provider-output fix. It is not: the closure is captured in `configResolved` and passed to Queue's provider build independently of `enforce`; the real consumer regression covers manual and generated Cloudflare bindings plus exclusion of Vercel output.
Next action: Keep the one-line ordering deletion, verify the focused consumer regression and relevant package checks, then review the exact base diff.

onmax:
@codex review

PR #775, opened 2026-07-21
fix(deploy): emit JavaScript for Deno Deploy

onmax:
Nitro's Deno preset emits its server bundle with a `.ts` extension, so Deno Deploy type-checks generated JavaScript and fails on Sharp's undeclared `@types/node` dependency before the app starts. ViteHub now emits the generated bundle as `server/index.mjs`, aligns its deployment plan, task, and create runner, and keeps a tiny computed-import `server/index.ts` shim for existing apps that retain the old entrypoint.

The generated-artifact regression runs Deno 2.9 checks through both entrypoints, exercises a real Linux x64 Sharp PNG transform through each path, and simulates an existing-app upload from the ignore-independent staging root with the native Sharp/libvips tree intact. Existing Nitro Rollup output options remain composed.

### DIFF
--- a/packages/internal/src/build/deno-runtime-packages.ts
+++ b/packages/internal/src/build/deno-runtime-packages.ts
@@ -332,7 +332,7 @@ try {
   await cp(sourceRoot, uploadRoot, { recursive: true })
 
   const common = ["--allow-node-modules", "--org", organization, "--app", app]
-  const creation = await run(["deploy", "create", ".", "--source", "local", "--do-not-use-detected-build-config", "--runtime-mode", "dynamic", "--entrypoint", "server/index.ts", "--working-directory", ".", "--region", region, ...common], uploadRoot)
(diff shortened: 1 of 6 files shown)


onmax:
> [!NOTE]
> **Direction verdict: revise**
>
> Emit Deno Deploy's canonical generated server entry as `server/index.mjs`, preserve all existing Rollup output options, and generate `server/index.ts` only as a compatibility shim.
>
> Evidence: the plan, runner, generated artifact, and Deno check/run regressions align on `.mjs`; exact-head Codex review identified that object-only composition discarded valid array-valued Rollup outputs. The required correction is to map array outputs while retaining every non-entry option, with focused regression coverage.
>
> The revision is applied on verified head `2ab2b4b4fb3747c7b6819562b1d81cd1500343e0`: object and array output tests pass, and base-relative spec review found no remaining findings. Validation artifacts match the current PR title/body intent.

onmax:
@codex review

PR #777, opened 2026-07-21
feat(workspace): preserve untracked GitHub files when publishing

onmax:
Adds a `deleteUntracked` opt-out to the GitHub Workspace publisher. Existing exact-mirror behavior remains the default; setting it to `false` publishes changed Workspace files without deleting remote-only paths under the configured root, including files retained from earlier snapshots.

Reconstructs the Quiver partial-mirror fix in source and documents the ownership trade-off. Focused publisher regressions cover default deletion, preservation during a changed-file update, and no-op detection when only remote-only paths differ.


onmax:
\n**Verdict: proceed**\n\nThe PR matches ViteHub's direction: this is a small, provider-specific Workspace publishing primitive that makes ownership boundaries explicit without changing the exact-mirror default.\n\nEvidence at head :\n- The API is a single opt-out on the existing GitHub Workspace publisher; it does not introduce an app-level workflow.\n- Default behavior remains deletion of remote-only paths, preserving the current exact-mirror contract.\n-  uses the existing remote-tree comparison and commit machinery while suppressing only deletions.\n- Focused tests cover changed-file publishing and the remote-only no-op case, and the README documents the ownership trade-off.\n- The branch is rebased on ; focused publisher tests and Workspace typecheck pass.\n

onmax:
@codex review

onmax:
@codex review

onmax:
@codex review

PR #781, opened 2026-07-21
fix(queue): support H3 2 runtime event types

onmax:
Queue's generated Cloudflare Nitro middleware now reads provider compatibility fields through a local runtime-event view, so H3 2 and Nitro 3 typechecking succeeds without changing event identity or provider behavior. The existing environment fallback order remains unchanged, and the typed event still enters Queue runtime state.

This adds generated-output coverage that compiles the emitted middleware against the workspace's Nitro 3 and H3 2 types and keeps Vercel middleware free of the compatibility escape hatch.


onmax:
> [!NOTE]
>
> **Head:** `e4cdbb66edc64de0ece68ea48e4a94d5f38170a3`
>
> **Intent:** Make generated Cloudflare Nitro queue middleware typecheck with H3 2 / Nitro 3 without changing runtime event identity, provider behavior, or environment fallback order.
>
> **Evidence:** The two-file diff is scoped to the generated middleware and its tests; the original `event` is still mutated and passed to `enterQueueRuntimeEvent`; every existing Cloudflare environment fallback remains in the same order; the compatibility view is emitted only for Cloudflare while Vercel output is explicitly checked; and the generated file is compiled against the workspace H3 2 / Nitro 3 types.
>
> **Conclusion:** This is a small provider-boundary compatibility repair aligned with ViteHub’s host-independent Queue runtime primitives. No overlapping implementation exists on current `main`, and the change preserves the external runtime contract.

onmax:
@codex review

PR #786, opened 2026-07-21
fix(blob): type generated Cloudflare serving middleware

onmax:
Generated Blob Cloudflare middleware now narrows Nitro's generic H3 event through a local structural carrier before reading the supported Cloudflare environment paths, and narrows the Worker environment at the same boundary. This preserves the existing lookup order while allowing Nitro 3 Cloudflare module builds to type-check without widening shared runtime APIs.

Adds a strict TypeScript regression that generates the real middleware through `hubBlob()` and compiles it with `cloudflare:workers` environment typed as `unknown`. The previous source fails on the generic event and request shapes; the updated generated source compiles.


onmax:
**Verdict: proceed**

Evidence:
- Exact evaluated head: `6caa215bf141ec6674bc54e27144efec9e6edf76`; intent matches the current title and body.
- The two-file change keeps Cloudflare-specific event structure local to generated Blob middleware and preserves the existing environment lookup order.
- The regression generates the real middleware and strictly compiles it with `cloudflare:workers` environment typed as `unknown`.
- The branch is reconciled with `main`; the focused generated-middleware regression, Blob type-check, and Blob build pass locally.

Required follow-through: obtain passing GitHub checks and a positive exact-head review.

onmax:
@codex review

PR #790, opened 2026-07-21
fix(deploy): prune incompatible Deno native packages

onmax:
Deno deployment output currently copies every installed optional native package and can recursively duplicate selected native closures. A Drop consumer with multiple supported architectures produced a 383 MB artifact that failed during Deno Deploy artifact preparation. An intermediate Linux build still retained both GNU and musl variants at 71 MB; hosted revision `mz5dpst907gm` uploaded successfully but failed during artifact preparation before user code ran.

Filter only optional package edges with npm-compatible os, cpu, and libc metadata for Deno Deploy's documented Linux x64/ARM64 GNU runtimes, hoist each selected native closure once, and preserve directly imported or required dependencies. The resulting exact Drop artifact is 36.8 MB, hosted revision `zjr47ymhm5eq` deployed successfully, and its full upload/optimization/stats E2E passed. The regression covers platform metadata semantics, generated artifact size and layout, and a real Deno Sharp transform.

Observed while validating vite-hub/drop#1.

### DIFF
--- a/packages/internal/src/build/deno-runtime-packages.ts
+++ b/packages/internal/src/build/deno-runtime-packages.ts
@@ -200,7 +200,7 @@ async function copyPackageToNodeModules(name: string, resolver: NodeJS.Require, 
       ? outputNodeModules
       : join(targetDir, "node_modules")
     await copyPackageToNodeModules(dependencyName, packageRequire, packageDir, dependencyNodeModules, copied, staged, {
-      hoistOptionalDependencies: options.hoistOptionalDependencies,
+      hoistOptionalDependencies: options.hoistOptionalDependencies && Boolean(packageJson.optionalDependencies?.[dependencyName]),
       includeOptionalDependencies: options.includeOptionalDependencies,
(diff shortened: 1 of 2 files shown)


onmax:
**Verdict: proceed**

Head `dcc0cb7343d6fce276bd01ba149e6110dde6a6fd` matches the PR intent and current `main`.

Evidence:
- The change remains confined to Deno deployment package staging and regression tests.
- Optional native edges are filtered with npm-compatible `os`, `cpu`, and `libc` constraints for the documented Deno Linux x64/ARM64 glibc targets.
- Direct dependencies remain staged even when their platform metadata is incompatible.
- Selected optional closures are hoisted once, while the follow-up repair keeps optionals reached through regular dependency edges version-isolated.
- Regression coverage now asserts both the hoisted native layout and the isolated regular-dependency layout; the intent is not already present on `main`.

onmax:
@codex review

onmax:
@codex review

PR #798, opened 2026-07-22
feat(sandbox): support function entrypoints with binary values

onmax:
Package Sandbox entrypoints can now default-export an async `(payload, context) => result` function. Generated registry types infer that function's payload and awaited result, use `unknown` for zero-argument functions, and retain the existing `SandboxPayload` fallback for non-function defaults.

Sandbox recursively stages `Blob` and `Uint8Array` leaves through invocation-local Box files, so application payloads and results no longer need base64 JSON wrappers. The transport uses collision-safe tagged values and confined integer sidecar IDs, while current top-level result exports remain compatible.

Drop proved the same 23,501-byte JPEG optimizing to 9,484 bytes across [Vercel](https://vitehub-drop-vcl-20260720-0204.vercel.app), [Cloudflare](https://drop.maximogarciamtnez.workers.dev), [Deno](https://vitehub-drop-deno-20260720-0204.onmax.deno.net), and [Netlify](https://vitehub-drop-provider-matrix-20260719.netlify.app); the Vercel and Cloudflare paths exercised Queue plus Sandbox binary input and output.

PR #796, opened 2026-07-22
feat(runtime): expose execution authority

onmax:
Defines an immutable, provider-independent `ExecutionAuthority` snapshot across filesystem, network, environment, credentials, processes, and isolation. Box providers and the local Agent harness declare their own authority; Box sessions, Workspace Sessions, Sandbox runners, and `AgentInspectionMetadata.config.driver.executionAuthority` expose the normalized result without inferring safety from executable selection. Missing or unprovable dimensions stay explicitly `unknown`, while processless surfaces report no execution.

Replaces Box's isolation-only metadata, adds first-party provider declarations and propagation coverage, and documents why structural command arguments and executable allowlists are dispatch controls rather than isolation boundaries.

## Stack

This PR is stacked on #794 (`agent/remove-browser-devtools`) and should remain based on that branch until the inspection transport lands. CLI authority inspection and composition validation can stack on this PR and consume `ExecutionAuthority` from `@vite-hub/runtime` through `AgentInspectionMetadata.config.driver.executionAuthority`.

PR #800, opened 2026-07-22
refactor!: standardize explicit results as error-first tuples

onmax:
Standardizes ViteHub’s deliberate non-throwing operation results on error-first tuples. Sandbox now returns `[SandboxError | null, value | undefined]`, Rate Limit drivers return `[Error | null, decision | undefined]`, and the local Sandbox resolver and controlled-curl parsers use the same ordering. Wrapper methods and object discriminants are removed, and public examples, playground callers, consumer fixtures, and type tests now destructure the error first.

Domain observations and externally owned records remain unchanged: `ShellAnalyzeResult` preserves command metadata for both allowed and rejected observations; command outputs, Standard Schema and AI SDK validation, lifecycle/status records, aggregate outputs, and Sandbox’s serialized subprocess envelope are ordinary domain or wire records rather than explicit operation results. Tuple aliases stay feature-owned instead of introducing a shared generic while Runtime and Sandbox dependency ownership is active in #796. #798 independently changes Sandbox entry transport and may require a small docs/test rebase.

Validated with Sandbox, Rate Limit, and Shell builds, typechecks, and full package suites. The Sandbox type test proves an `if (error)` guard narrows the destructured value, while the consumer fixture and Rate Limit published-package fixture validate the public contracts.

### DIFF
--- a/docs/content/docs/server-primitives/rate-limit.md
+++ b/docs/content/docs/server-primitives/rate-limit.md
@@ -138,7 +138,7 @@ const decision = await limiter.consume({ key: 'demo' })
 
 Every direct limiter exposes its resolved `policy` and the provider capabilities that affect enforcement and deployment. The memory driver is process-local and intended for development, tests, and known single-process hosts.
 
-Custom drivers return `{ unavailable: true, cause }` only for expected operational outages that the declared failure policy should govern. Configuration, provider-contract, and implementation defects should throw normally.
+Custom drivers return `[null, result]` after consuming the counter and `[error, undefined]` only for expected operational outages that the declared failure policy should govern. Configuration, provider-contract, and implementation defects should throw normally.
 
 ## Deploy to Cloudflare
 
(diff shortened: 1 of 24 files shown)


onmax:
**Verdict: proceed**

**Head:** `a0526a3910a0b3b7b131980abee1819a2a6b3fea`

**Intent:** Standardize deliberate non-throwing operation results as error-first tuples while leaving domain observations, wire records, and externally owned validation results unchanged.

**Evidence:**
- The branch is one focused commit directly on current `main` at `0bf37a40cbdd0b59a2b7c44ad6c8896ec9dbf8d4`.
- Sandbox and Rate Limit expose feature-owned success/error tuple unions; no shared generic or new cross-package dependency is introduced.
- Local Sandbox resolution and controlled-curl parsing follow the same explicit-result ordering without changing their external behavior.
- Public documentation, playground callers, consumer fixtures, runtime tests, and type tests are updated together.
- Exact-head verification passes: Sandbox build/publint and 203 tests; Rate Limit build/publint, 71 tests, and declaration checks; Shell build/publint and 21 tests; `git diff --check` is clean.
- Exact-base review found no actionable or unrelated changes, and repository-wide searches found no stale result-wrapper or unavailable-discriminant consumers.

**Strongest challenge:** Error-first tuples must retain precise narrowing and failure causes while malformed provider responses and programmer errors continue to throw rather than being misclassified as expected operational failures.

**Next action:** Require passing exact-head checks and a positive exact-head Codex review before merge.

onmax:
@codex review

PR #802, opened 2026-07-22
test(rate-limit): add Cloudflare provider conformance

onmax:
Cloudflare Rate Limit now participates in the shared Vite playground provider conformance. `vp run rate-limit:e2e --mode local|live --provider cloudflare --url ...` verifies generated binding resolution, explicit keys, default client-address resolution, and eventual best-effort `429` enforcement; the local provider runner and nightly Live Smoke both call the same package-owned suite.

Vercel is explicitly unsupported in the composer, runner, docs, and test routes, so the playground cannot silently fall through to process memory. The local Cloudflare runner also disables Wrangler containers, matching the existing Sandbox-live-only boundary and allowing the declared local primitive matrix to run without Docker.

Validated with the full Cloudflare local provider suite, Cloudflare and Vercel provider-output tests, Rate Limit package tests including the published-types consumer, launch-trust docs tests, lint, and a direct Vercel bridge check returning `501`. No provider deployment was performed; the deployed path is wired into the existing Live Smoke workflow.

Once this merges, downstream apps such as Drop can remove generic Rate Limit/provider loops and keep only product composition and media-policy smoke checks.


onmax:
**Verdict: `proceed`**

Head: `62a4f1473595f2acce1dc5d652037791463fb1a1`
Intent: package-owned Cloudflare Rate Limit conformance shared by local and live provider runs, with explicit Vercel non-support.

Evidence:
- The playground reuses existing Rate Limit discovery, Cloudflare binding generation, and runtime configuration primitives.
- The same `packages/rate-limit/test/e2e.mjs` suite is invoked by the local Cloudflare runner and nightly Live Smoke.
- Vercel composition disables Rate Limit and test routes return 501 without a Cloudflare environment, preventing silent memory fallback.
- The branch is directly based on current `origin/main`; all current GitHub checks pass.

Risk retained: no provider deployment was performed for this head; the credentialed nightly workflow is the live proof boundary.

onmax:
@codex review

PR #805, opened 2026-07-23
Accept H3 events in requireRateLimit

onmax:
Narrows `RateLimitRequestEvent` to the `req.headers.get` surface that every accepted event must provide, so H3 events with host-owned request contexts can pass directly to `requireRateLimit`. H3's `getRequestIP()` still owns optional client-address fallbacks at runtime; this removes ViteHub's incompatible redeclaration of those shapes without changing enforcement or provider behavior.

Adds focused type coverage derived from the current `H3Event` with a host-specific request context. The full Rate Limit package passes all 71 tests with zero type errors, including build, publint, and the published-types consumer; package typecheck, focused lint, and `git diff --check` also pass.

### DIFF
--- a/packages/rate-limit/src/types.ts
+++ b/packages/rate-limit/src/types.ts
@@ -25,9 +25,7 @@ export interface RequireRateLimitOptions extends RateLimitPolicy {
 
 export interface RateLimitRequestEvent {
   readonly req: {
-    readonly context?: { clientAddress?: string }
     readonly headers: { get: (name: string) => string | null }
-    readonly ip?: string
   }
 }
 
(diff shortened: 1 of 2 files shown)


onmax:
> [!NOTE]
> **Direction verdict: accept**
>
> Head `0c43b554823be3e18ef858789f36e9ca3c2c0be8` matches the stated intent: it removes only the incompatible optional `req.context` and `req.ip` redeclarations from `RateLimitRequestEvent`, preserving the `req.headers.get` surface consumed by H3 client-address resolution. Focused type coverage now uses an `H3Event` with a host-owned request context, directly proving the compatibility gap. The two-file, 8-addition/4-deletion diff aligns with ViteHub’s provider-agnostic and obvious-API direction and introduces no unrelated scope.

onmax:
@codex review

PR #807, opened 2026-07-23
feat(vite-hub): infer deployment identity from package name

onmax:
Adds one typed ViteHub deployment identity, resolved from `vitehub.name`, the legacy `VITEHUB_DEPLOYMENT_NAME`, the nearest `package.json` name, or the Vite root in that order. Differing explicit config and environment values now fail instead of allowing resource names to diverge.

Cloudflare uses the identity to default the Worker, R2 bucket, Queue prefix, Rate Limit namespace, and Sandbox name. Explicit Wrangler Worker names and Blob bucket, driver, or store options remain physical overrides, while the full provider-neutral identity and its source are recorded in `deployment.json`.

This also makes Blob consistent with the root composer's positive opt-in contract: omission or `false` disables it, `blob: true` enables the preset store, and an options object enables and configures it.

## Verification

- frozen workspace install
- `vite-hub` dependency build, package build, publint, 63 tests, and type tests
- focused deployment-manifest tests: 12 passed
- workspace package and docs typechecks
- docs tests: 39 passed
- production docs build
- Cloudflare consumer build without an env mutation or explicit Worker/bucket names: 2 passed, 1 platform-specific Deno test skipped
- focused lint and `git diff --check`


onmax:
**Verdict: proceed**

Head: `1d69c3b0d86a57dfb3e23702669caa6213165659`

Intent: establish one provider-neutral deployment identity, infer it from the nearest package name with explicit overrides, apply it to Cloudflare resource defaults, record it in deployment output, and make Blob follow the root composer’s opt-in contract.

Evidence:
- This strengthens a ViteHub runtime/deployment primitive rather than adding an app-level workflow.
- One inspectable identity shared across provider output and capability defaults makes the obvious agent/developer assumption true and keeps provider plumbing behind ViteHub language.
- Explicit physical-resource overrides remain authoritative, preserving composability and provider control.
- Blob’s positive opt-in behavior aligns it with the root composer’s existing capability contract.
- Exact-head repairs derive valid generated R2 bucket names and carry disabled Blob state into generated Agent runtime contexts.

onmax:
@codex review

onmax:
@codex review

onmax:
@codex review

onmax:
@codex review

onmax:
@codex review

onmax:
@codex review

PR #809, opened 2026-07-23
fix(blob): use cached headers for served blobs

onmax:
Generated Blob serving routes now use Nitro's headers-only cached handler. Responses without a configured `Cache-Control` receive public immediate-revalidation headers (`public, max-age=0, s-maxage=0`), while `blob.serve.headers` still overrides that default.

Headers-only mode is required because Nitro/ocache's full response-cache path reads `Response.text()` and stores a string body, which can change arbitrary Blob bytes. The regression bundles and executes the generated route, then proves invalid UTF-8 bytes remain exact with both default and configured cache policies. This does not add stored response caching or automatic ETag-based `304` behavior.

## Verification

- Blob package typecheck
- generated-route and binary cache-header tests: 3 passed
- Blob package suite: 144/147 passed; the remaining three standalone-artifact tests fail before the serving route while resolving malformed temporary-directory imports

### DIFF
--- a/packages/blob/src/vite.ts
+++ b/packages/blob/src/vite.ts
@@ -175,12 +175,13 @@ function renderBlobServeRouteHandler(serve: BlobServeConfig, importBase = blobPa
   const headers = serve.headers && Object.keys(serve.headers).length > 0 ? serve.headers : undefined
   return [
     `import { blob } from '${importBase}'`,
-    `import { createError, defineEventHandler, getRouterParam${headers ? ", removeResponseHeader, setResponseHeaders" : ""} } from 'h3'`,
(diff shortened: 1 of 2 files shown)


onmax:
**Verdict: proceed**

The exact prepared head `3a5af7b2872b813aecfd152186994fdadac0ec5c` matches the PR intent. The change uses Nitro’s existing `defineCachedHandler` primitive in `headersOnly` mode, avoids the body-serialization path that can corrupt arbitrary Blob bytes, preserves explicit `blob.serve.headers`, and adds an executed binary-response regression for both default and configured cache policies. Current `main` does not already contain an equivalent change, the diff is limited to the generated Blob route and its tests, and the PR is cleanly mergeable.

onmax:
@codex review

PR #811, opened 2026-07-23
refactor(runtime): return Blob and KV operation results

onmax:
## Summary

- Return error-first tuples from every async Blob and KV operation, including provider resolution, with typed ViteHub errors that retain the operation, selected store, and raw cause.
- Keep invalid input and unknown-store configuration throwing, preserve generated Blob 404 responses, and retain the cached serving handler from #809.
- Migrate first-party runtimes, generated fixtures, examples, playground routes, docs, and the Schedule KV adapter to the new contract.

## Validation

- Blob: 151 tests passed, no type errors.
- KV: 49 tests passed, no type errors.
- Schedule: 195 tests passed, no type errors.
- Database: 78 tests passed, no type errors.
- Queue: changed composed Blob runtime case passed; the full local suite still has an unrelated unchanged shared-output assertion failure.
- All workspace package typechecks, docs typecheck, lint, and `git diff --check` passed.
- Applied the built Blob and KV packages through `pnpm patch` in an isolated Drop copy; its Cloudflare production build and `tsc --noEmit` passed with no application `try/catch` remaining.


onmax:
**Verdict: proceed**

Head `e9f383b9a3e0a1851f5d284e3f01ca1ccc62fd88` matches the PR intent after a clean rebase onto `main` at `84d4fa1`. The change makes Blob and KV async operations return explicit typed error-first results, preserves throwing for invalid input/configuration, carries operation/store/cause diagnostics through the existing `ViteHubError` primitive, migrates first-party consumers, and documents/tests the breaking contract. This aligns with the repository direction toward obvious, provider-agnostic, inspectable server primitives and its explicit preference for the final API over backward compatibility.

Evidence: the public Blob/KV interfaces and runtime wrappers use one consistent result contract; named-store/provider failures are classified; Schedule adapts the changed KV boundary without leaking it into Schedule internals; docs, fixtures, generated-output assertions, capability adapters, and type tests are updated. The 62-file footprint is migration fallout from the public breaking API rather than unrelated product scope. Rebase reconciliation was conflict-free; lint and workspace typecheck pass; Blob (151) and KV (49) package tests pass with no type errors.

onmax:
@codex review

PR #816, opened 2026-07-24
fix(blob): keep R2 fallback in HTTP mode

onmax:
Cloudflare R2 fallback now passes only HTTP adapter options to Files SDK, so a configured binding name cannot select binding mode when the runtime binding is absent. Supported `defaultUrlExpiresIn` and `publicBaseUrl` values remain intact.

This fixes the residual fallback mismatch after #767, where native-versus-HTTP selection moved into ViteHub but the HTTP call still received ViteHub’s `binding`, `driver`, and `bucketName` fields. The runtime regression asserts the exact provider option object and keeps the worker-environment credential cases aligned.


onmax:
**Verdict: proceed**

- Exact head: `64b31da197cbd12ec83935e0e432d699ccf90a24`
- Intent: keep the missing-binding Cloudflare R2 fallback in HTTP mode by passing only supported HTTP adapter options to `files-sdk/r2`.
- Evidence: current `origin/main` is an ancestor of the head; ViteHub already owns native-versus-HTTP selection in runtime storage; the diff is limited to the explicit provider mapping and its fallback regression.
- Required correction: none. Preserve `defaultUrlExpiresIn` and `publicBaseUrl`, then verify the package and review the exact base-relative diff.

onmax:
@codex review

PR #835, opened 2026-07-25
fix agent chat adapter version typing

onmax:
## What changed

- Add a structural `AgentChatPlatformAdapter` boundary that preserves the adapter methods ViteHub uses while keeping Chat SDK version-bound `Message` and `ChatInstance` values opaque.
- Pin ViteHub's owned `chat` dependency to 4.34.0 so the compatibility test exercises a real cross-version install.
- Add a packed consumer test with `@chat-adapter/telegram@4.35.0`, including `allowedUserIds`, under strict peer checking.

## Why

Chat SDK adapters own an exact `chat` dependency. A 4.35 adapter was rejected by ViteHub's 4.34 `Adapter` type because `Message` contains private state, even though the runtime adapter surface is compatible. Weakening the whole resolver to `any` hides too much; this keeps the stable adapter surface typed and isolates only the version-bound values.

## Validation

- `pnpm install --frozen-lockfile --ignore-scripts`
- `pnpm --filter @vite-hub/agent test` — 62 files, 1,449 tests
- packed Telegram 4.35 published-type consumer under `--strict-peer-dependencies`
- `pnpm exec vp run test:consumer`

## Dependency boundary

A consumer that also installs Workflow 4 still cannot use Telegram 4.35 with strict peer checking because Chat 4.35 requires `workflow@^5.0.0-beta.35`. That Workflow 5/provider-output migration is a separate dependency seam; this PR does not suppress or weaken that peer contract.

### DIFF
--- a/fallow.toml
+++ b/fallow.toml
@@ -108,6 +108,8 @@ ignoreDependencies = [
   "@vercel/blob",
   "@vercel/queue",
   "chat-state-cloudflare-do",
+  # Installed only inside the packed Agent Chat Platform Adapter consumer fixture.
+  "@chat-adapter/telegram",
   # Package builds and generated provider output reference these without source imports.
   "@aws-sdk/client-s3",
   "@aws-sdk/lib-storage",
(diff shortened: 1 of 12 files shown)


onmax:
**Verdict: proceed**

Head fb0cb3e48baede8053dc7c658ca569962ce18984 matches the PR intent: it moves a downstream Chat SDK version seam into ViteHub's Agent Chat Platform Adapter boundary, preserves the runtime methods ViteHub uses, and keeps only version-private Chat SDK values opaque. The packed Telegram 4.35 consumer exercises the public package contract under strict peer checking. The follow-up only registers that isolated fixture dependency with repository dead-code analysis. This remains a focused upstream primitive consistent with ViteHub's provider-agnostic, composable Agent Definition direction.

onmax:
@codex review

PR #833, opened 2026-07-25
feat(database): run database definitions directly

onmax:
## Summary

- make `defineDatabase()` return a typed Drizzle runtime while preserving its definition metadata
- select local or hosted adapters through the internal `vitehub-hosted` export condition
- use `schema` as the definition source for discovery, generated exports, examples, and published types
- propagate the hosted condition through Database, Nuxt, Cloudflare, Vercel, composed Blob output, and the provider composer
- require Named Database runtime names to match their discovered identity
- pin the supported Chat version so clean consumers retain Workflow v4 compatibility

## Validation

- `vp run -t @vite-hub/database#build`
- `pnpm --filter @vite-hub/database typecheck`
- `pnpm --filter @vite-hub/blob typecheck`
- `vp lint packages/database packages/blob`
- `vp exec fallow dead-code --summary --format markdown --fail-on-issues`
- `pnpm --filter @vite-hub/database exec vp test` (88/88)
- `pnpm --filter @vite-hub/blob exec vp test test/vite-output.test.ts` (33/33)
- `vp test --config test/output/vitest.config.ts test/output/vercel.test.ts` (6/6)
- `vp run test:consumer` (1/1)

PR #837, opened 2026-07-25
fix(blob): select hosted Database definitions

onmax:
Blob-owned Cloudflare and Vercel provider output now includes the `vitehub-hosted` resolution condition when a prepared Database runtime is composed. Standalone Blob output keeps its existing resolution conditions.

Database PR #833 adds conditional local and hosted definition runtimes, but Blob becomes the final provider-output owner when the primitives compose. Without this condition, the final bundle selects the local definition branch and Cloudflare rejects its `node:fs` and `node:path` imports. Focused conditional-export regressions prove that both Blob-owned outputs select the hosted branch.

Validated with the full `@vite-hub/blob` build and test command (153/153), package typecheck, scoped lint, and `git diff --check`. Refs #833.

### DIFF
--- a/packages/blob/src/internal/vite-build.ts
+++ b/packages/blob/src/internal/vite-build.ts
@@ -413,7 +413,13 @@ function createCloudflareOutput(blob: BlobModuleOptions | ResolvedBlobModuleOpti
         ...(databaseRuntime ? { "@vite-hub/database/drizzle": databaseRuntime } : {}),
       },
       banner: `// ${cloudflareBlobWorkerMarker}`,
-      conditions: ["workerd", "worker", "browser", "default"],
+      conditions: [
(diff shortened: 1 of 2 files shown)


onmax:
> [!NOTE]
>
> Head `0888a06cbcf66c817d9392469c4db490cbfb8fb0` matches the PR intent. Blob owns the final Cloudflare and Vercel bundle when a Database runtime is composed, so propagating `vitehub-hosted` at that boundary makes the hosted conditional export reachable without changing standalone Blob resolution. The two focused output tests cover both owned providers. Repository evidence: current `main` still lacks these conditions, while open Database PR #833 depends on this commit and documents the same provider-composition contract. Verified after rebasing onto `3d59a45`: Blob build and 153/153 tests, package typecheck, scoped lint, and `git diff --check` pass.

onmax:
@codex review

PR #839, opened 2026-07-25
fix(queue): exclude Cloudflare output from Vercel static

onmax:
Stops Queue Vercel output from copying the preserved default Cloudflare output subtree into `.vercel/output/static` when both outputs share `dist`. The cleanup is scoped to a nested output with a source `wrangler.json`, so shared Cloudflare config remains in its owner output and ordinary static assets still deploy.

The existing shared-output regression now passes, along with the full Queue and provider-output suites.

### DIFF
--- a/packages/internal/src/build/deployment-output.ts
+++ b/packages/internal/src/build/deployment-output.ts
@@ -218,7 +218,7 @@ async function copyVercelClientOutput(rootDir: string, clientDir: string, static
     && !isAbsolute(outputRelativePath)
 
   await copyClientOutput(clientDir, staticOutputDir)
-  if (!excludesCloudflareOutput) return
+  if (resolve(clientDir) === resolve(staticOutputDir) || !excludesCloudflareOutput) return
 
   try {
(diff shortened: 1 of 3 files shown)


onmax:
**Verdict: proceed**

Head `01db0bd8bf54dbcad705fdcaa93b98aa8e40c2e9` matches the PR intent. The diff moves the exclusion to the shared Vercel static-copy boundary in `packages/internal`, where every provider writer applies it, and removes the Queue-local cleanup that a later writer could undo. The exclusion remains limited to a nested default Cloudflare output containing its source `wrangler.json`, and it skips cleanup when the resolved static target aliases the client source. Regression coverage proves repeated Vercel writes and an aliased custom target preserve ordinary static assets and the source Cloudflare output while excluding only a distinct copied subtree. No unrelated scope or obsolete direction-validation body section remains.

onmax:
@codex review

onmax:
@codex review

onmax:
@codex review

PR #843, opened 2026-07-26
fix(blob): unpublish bundled Vercel driver

onmax:
Removes `@vite-hub/blob/drivers/vercel-bundled` from the package export map so ViteHub no longer presents its hosted dependency-closure strategy as a selectable Blob Driver. The implementation remains built because ViteHub selects it internally when `vercel-blob` runs in hosted Vercel output; the playground E2E source import is unchanged.

Adds a focused manifest regression. The Blob build and publint, package typecheck, public-export test, Nitro selector tests, isolated Vercel output test, and the Vite E2E Vercel build pass.

A full Blob sweep reached 150 passing tests; three temporary-output path cases failed under the linked dependency worktree, and the clean-consumer install timed out without network. The affected bundled-runtime and Vercel output paths were rerun directly and pass.

### DIFF
--- a/packages/blob/package.json
+++ b/packages/blob/package.json
@@ -51,7 +51,6 @@
     "./drivers/supabase": "./dist/drivers/supabase.js",
     "./drivers/uploadthing": "./dist/drivers/uploadthing.js",
     "./drivers/vercel": "./dist/drivers/vercel.js",
-    "./drivers/vercel-bundled": "./dist/drivers/vercel-bundled.js",
     "./ensure": "./dist/ensure.js",
     "./errors": "./dist/errors.js",
     "./runtime/cloudflare-vite": "./dist/runtime/cloudflare-vite.js",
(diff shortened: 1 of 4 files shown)


onmax:
**Verdict: proceed**

The exact-head change at `1f83096671001d2c29637b3e8fcd00618b291ae2` matches the PR intent and ViteHub direction: it removes a hosted, internal dependency-closure implementation from the public Blob Driver selection surface while preserving ViteHub’s provider-agnostic `vercel` driver and internal hosted-output behavior. The package build now explicitly filters the internal entry from generated exports, while still building it for hosted output. Regression evidence covers the generated manifest and internal runtime export contract; Fallow, Blob build/publint, focused manifest tests, and Blob typecheck pass.

onmax:
@codex review

onmax:
@codex review

PR #854, opened 2026-07-27
fix(agent): bundle AI SDK for Cloudflare deployments

onmax:
`@vite-hub/agent` now loads the AI SDK through a literal lazy import, so Nitro can include the dependency in Cloudflare's module graph instead of leaving a bare runtime import.

A production Worker failed before its Agent Driver or model ran with `No such module "_libs/@vite-hub/ai".` Applying this package change in the downstream consumer made Nitro emit a resolvable AI module. The regression builds a real Nitro `cloudflare-module` fixture, verifies the generated route points to an existing relative module, and imports that module to confirm it exposes `generateText`.

Validated with the Agent build and publint, Agent typecheck, Agent config tests, and the deployment-preset suite.

### DIFF
--- a/packages/agent/src/internal/ai-sdk-runtime.ts
+++ b/packages/agent/src/internal/ai-sdk-runtime.ts
@@ -1,5 +1,3 @@
-const aiSdkPackageName = "ai"
-
 export async function loadAiSdk(): Promise<typeof import("ai")> {
-  return await import(/* @vite-ignore */ aiSdkPackageName) as typeof import("ai")
+  return await import("ai")
 }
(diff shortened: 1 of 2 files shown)


onmax:
**Verdict: proceed**

- Head: `8a98db0e3ff271e6147b0fe0391702082d3e074f`
- Intent: make the Agent Package's lazy AI SDK load statically discoverable so Nitro includes it in Cloudflare Provider Output.
- Evidence: the change replaces the opaque `@vite-ignore` variable import with a literal lazy `import("ai")`; `ai` is a declared Agent Package dependency; the regression builds Nitro's `cloudflare-module` preset, resolves the emitted relative module, and imports it to verify `generateText`.
- Direction: this preserves the host-neutral Agent API and keeps Nitro/Cloudflare details in package-owned generated output, matching the repository's Agent-first and provider-agnostic boundaries.
- Base: exact parent is current `origin/main` at `7c1fb74`; no reconciliation or unrelated scope is present.

onmax:
@codex review

PR #857, opened 2026-07-27
fix(vite-hub): align Nuxt generated handler roots

onmax:
Anchors Agent-generated Nitro handlers to the resolved Vite root that owns `.vitehub`. Nuxt projects with a separate `srcDir` now register `app/.vitehub/agent/chat-webhook-route.ts` instead of asking Nitro to resolve the same relative path from the project root.

Agent continues to own both generation and registration, so the Nuxt adapter does not rewrite consumer-defined or sibling ViteHub handlers. The regression runs the real `vite-hub/nuxt` and Agent config hooks with distinct `rootDir`, `srcDir`, and `serverDir`; it failed with the relative handler before this change and passes with the source-root-anchored path.

Validated with the full Agent suite (64 files, 1,469 tests), the full `vite-hub` suite (8 files, 71 tests), both package typechecks, build and publint through the package test pipelines, `git diff --check`, and independent standards/spec reviews. A disposable Nuxt 4.4.8 consumer on the established Vite 8.1.5 graph completed a production Nitro build, generated `app/.vitehub/agent/chat-webhook-route.ts`, and bundled its project-root Agent import successfully.

### DIFF
--- a/packages/agent/src/vite.ts
+++ b/packages/agent/src/vite.ts
@@ -1808,25 +1808,26 @@ export function hubAgent(options?: AgentModuleOptions): AgentVitePlugin {
       agent = config.agent ?? agent
       const resolved = normalizeAgentOptions(agent)
       serverDirs = (config as typeof config & { [VITEHUB_SERVER_DIRS]?: string[] })[VITEHUB_SERVER_DIRS] ?? serverDirs
-      const hasHostedAgents = Boolean(resolved && hasHostedAgentDefinitions(resolve(config.root || process.cwd()), serverDirs))
+      const root = resolve(config.root || process.cwd())
+      const hasHostedAgents = Boolean(resolved && hasHostedAgentDefinitions(root, serverDirs))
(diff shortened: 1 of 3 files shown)


onmax:
**Verdict: proceed**

Head: `61dda832413740764c0dae49cdb31be1d8e3e10c`
Intent: resolve Agent-generated Nitro handler paths from the Vite root that owns `.vitehub`, including Nuxt projects with a distinct `srcDir`.
Evidence: rebased cleanly onto `main` at `9639e51`; the implementation anchors only Agent-owned generated webhook and Discord handlers to resolved `config.root`, which is also the generation root. The regression exercises the real Nuxt and Agent hooks with distinct `rootDir`, `srcDir`, and `serverDir`; existing provider tests cover all registered generated handlers. Verified with Agent 64 files/1,472 tests, vite-hub 8 files/71 tests, both typechecks, package builds/publint, and `git diff --check`. Base-diff review found no actionable findings.

onmax:
@codex review

PR #868, opened 2026-07-29
feat(database): support remote D1 during development

onmax:
Local Database Definitions now honor the existing `cloudflare.http` transport during development, so both direct `defineDatabase()` use and generated `@vite-hub/database/drizzle` queries can reach remote Cloudflare D1. The active binding still wins on Cloudflare, omitting `cloudflare.http` keeps local SQLite unchanged, and selecting HTTP without a database id fails before any local fallback.

This keeps provider selection inside the Database Definition instead of adding `database.remote` or a process-global mode flag. The shared adapter now treats explicit `cloudflare.http` as the single transport gate across runtimes, which lets Calories remove its downstream patch after moving the opt-in beside its Cloudflare database id.

A Vite development regression loads the generated Drizzle runtime and proves a query reaches an authenticated D1-compatible HTTP endpoint with the expected URL, authorization, SQL body, and returned rows. Direct-definition, missing-id, and local-SQLite regressions cover the other runtime paths. The full Database suite passes 96 tests with no type errors; Database typecheck, focused lint, docs typecheck, package build/publint, and diff checks also pass.

### DIFF
--- a/docs/content/docs/server-primitives/database.md
+++ b/docs/content/docs/server-primitives/database.md
@@ -120,7 +120,7 @@ A project uses either one Default Database or a set of Named Databases. Do not m
 | `connection.authToken` | `DatabaseConfigValue` | No | Hosted database auth token. |
 | `cloudflare.binding` | `string` | No | D1 binding. Defaults to `DB` for Default Database and `DB_<NAME>` for Named Databases. |
 | `cloudflare.databaseId` | `DatabaseConfigValue` | No | D1 database id. |
-| `cloudflare.http` | `true \| { url, authToken }` | No | Explicitly selects authenticated D1 raw HTTP access for hosted runtimes. `true` uses Cloudflare's API; an object selects a compatible proxy. |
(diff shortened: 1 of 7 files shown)


onmax:
> [!NOTE]
> **Direction verdict: `revise` (applied)** for head `d6cb6df83cb461e6030766f3ea9db0606832a8f0`.
>
> Intent: treat `DatabaseDefinition.cloudflare.http` as the explicit D1 HTTP transport gate across shared Drizzle runtimes, while preserving active binding precedence and SQLite/libSQL behavior when absent.
>
> Evidence: rebased onto current `main` at `98d72c4538000e24bb24680b14e4ff16847e793a`; explicit HTTP fails fast when `cloudflare.databaseId` is unavailable, addressing the P1 fallback finding. Exact-head local verification passes 96 Database tests with no type errors, Database typecheck, focused lint, and diff checks; prior docs typecheck and package build/publint pass on identical PR content. All review findings are addressed.
>
> Required correction: complete. Await exact-head GitHub checks and Codex review.

onmax:
@codex review

onmax:
@codex review

onmax:
@codex review

PR #869, opened 2026-07-29
fix(agent): type generated lazy chat state

onmax:
Generated Nitro and Deno Agent routes now give lazily initialized libSQL Chat state an explicit `ReturnType<typeof createLibsqlAgentState> | undefined`. Netlify `.mjs` keeps the plain JavaScript declaration, so the shared generator remains valid across hosts.

The regression compiles Cloudflare-native and libSQL-backed Nitro output under strict TypeScript, asserts the typed Deno output, and syntax-checks generated Netlify JavaScript. Before the source fix, the libSQL case reproduced TS7034 and TS7005; after it, all 180 provider tests, Agent typecheck, and Agent build/publint pass.

This upstreams only the downstream generated-state typing hunk; unrelated runtime patch behavior remains out of scope.

### DIFF
--- a/packages/agent/src/vite.ts
+++ b/packages/agent/src/vite.ts
@@ -943,7 +943,7 @@ function generatedCloudflareChatStateHelper(): string[] {
   ]
 }
 
-function generatedLibsqlChatStateHelper(state: GeneratedLibsqlAgentStateOptions): string[] {
+function generatedLibsqlChatStateHelper(state: GeneratedLibsqlAgentStateOptions, typescript = false): string[] {
   const { authTokenEnvName, durableUrlRequired, ephemeralHosting, ...stateOptions } = state
   const configuredAuthTokenOption = authTokenEnvName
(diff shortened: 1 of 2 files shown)


onmax:
**Verdict: proceed**

Generated Nitro and Deno TypeScript should explicitly type the lazy libSQL chat state as `ReturnType<typeof createLibsqlAgentState> | undefined`, while generated Netlify `.mjs` must retain its plain JavaScript declaration.

Evidence for head `74e7be3a657b5749bf0a6b0dc6f26c1eb345936d`:
- the type derives from the factory already imported by both TypeScript outputs;
- the shared helper keeps runtime initialization unchanged;
- all 180 provider tests pass, Agent typecheck passes, and the Agent dependency build/publint passes;
- the two-axis review against `main` found no standards or spec issues.

All four direction lenses recommend proceed. The only residual risk is future misuse of the boolean output-language flag; current call sites are adjacent and both syntax paths are covered.

onmax:
@codex review

PR #875, opened 2026-07-30
fix(blob): resolve generated Nitro handlers from the project root

onmax:
## Summary

- resolve Blob's generated Nitro plugin, Cloudflare middleware, and serving handler from the final Vite project root
- replace relative or already-resolved registrations with one canonical path so repeated configuration stays idempotent
- cover a consuming root that differs from `cwd`, including mixed relative and absolute registrations

## Why

Nuxt can materialize both workspace-root and app-root `.vitehub` trees. Relative Nitro registrations can silently select the stale workspace runtime even after Blob generates the correct app-root runtime, so the final Vite hook now anchors every generated registration to `config.root`. The early `config` hook remains relative because its root is not final yet.

## Verification

- `pnpm --filter @vite-hub/blob exec vp test test/vite.test.ts -t 'registers Nitro runtime setup|resolves generated Nitro registrations|facade declare Nitro ownership|final Nitro ownership|does not yield Cloudflare output|removes an early R2 contribution|masks credentials'`
- `pnpm --filter @vite-hub/blob build`
- `pnpm --filter @vite-hub/blob typecheck`
- `pnpm exec vp exec oxlint packages/blob/src/vite.ts packages/blob/test/vite.test.ts --ignore-path .gitignore`
- `git diff --check`

### DIFF
--- a/packages/blob/src/vite.ts
+++ b/packages/blob/src/vite.ts
@@ -112,21 +112,32 @@ function blobServeNitroRoute(serve: BlobServeConfig): string {
   return `${normalizeNitroRoute(serve.route).replace(/\/+$/, "")}/**`
 }
 
-function mergeNitroBlobConfig(value: unknown, serve: BlobServeConfig | undefined, cloudflare: boolean): Record<string, unknown> {
+function mergeNitroBlobConfig(value: unknown, serve: BlobServeConfig | undefined, cloudflare: boolean, root?: string): Record<string, unknown> {
   const nitro = cloneNitroConfig(value)
-  const plugins = Array.isArray(nitro.plugins) ? [...nitro.plugins] : []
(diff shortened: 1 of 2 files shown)


onmax:
**Verdict: proceed**

Exact head `725d272ecd43c7645da5e7a5f344c63f8ba3d05d` matches the PR intent and repository direction.

Evidence:
- Scope remains limited to `packages/blob/src/vite.ts` and `packages/blob/test/vite.test.ts`.
- Final `configResolved` registrations are canonicalized against `config.root`, while the early `config` hook remains relative before the root is final.
- Canonicalization now removes generated registrations already resolved against either the stale workspace root or the final app root.
- Regression coverage uses roots distinct from `process.cwd()` and proves repeated mixed relative/stale/final registration is idempotent.
- Focused tests, Blob build/typecheck, oxlint, and diff checks pass.

onmax:
@codex review

PR #878, opened 2026-07-30
fix(agent): materialize lazy sources for harness sessions

onmax:
## What changed

Harness-backed agents now pass each filesystem-backed lazy source mount into workspace session preparation. Request-only sources remain excluded, while non-lazy sources retain their existing probe-path behavior. The new provider-error logging regression test also uses a unique chat delivery ID so full-suite deduplication cannot skip its error path.

## Why

Harness sessions receive a physical filesystem snapshot. Selecting only probe paths meant lazy sources without probe keys were omitted entirely, and lazy sources with probe keys exposed only those probes rather than the complete source. Codex therefore could not inspect repositories that the agent workspace declared as available.

The provider test isolation repair is required by the rebased base: its delivery ID duplicated an earlier manual-delivery test, so the Chat SDK correctly deduplicated it during the full suite and returned before exercising the provider error.

## Validation

- `pnpm --filter @vite-hub/agent exec vp test test/workspace.test.ts` — 164 tests passed
- `pnpm --filter @vite-hub/agent exec vp test test/providers.test.ts` — 193 tests passed

### DIFF
--- a/packages/agent/src/harness-agent.ts
+++ b/packages/agent/src/harness-agent.ts
@@ -227,6 +227,9 @@ function workspaceSourceHarnessPaths(context: AgentAdapterRunContext): string[] 
   const sources = context.workspaceDefinition?.sources
   if (!sources) return []
   return normalizeWorkspaceSourcesMetadata(sources).flatMap((source) => {
+    if (source.materialize === "lazy" && !source.requestOnly) {
+      return [source.mountPath]
+    }
     if (source.probeKeys?.length) {
       return source.probeKeys.map(key => [source.mountPath, key].filter(Boolean).join("/"))
     }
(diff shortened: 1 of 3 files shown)


onmax:
**Verdict: proceed**

Head: `372282504503b02aa8f8cf8a748bc6ba7dda26ca`

Evidence: The PR intent is to make filesystem-backed lazy Sources available in Harness Workspace Sessions. The implementation uses the existing normalized Source metadata and session path-scoping primitive, promotes non-request-only lazy Sources from probe paths to their complete mount path, preserves build-source probe behavior, and adds focused regression expectations for lazy Sources with and without probes. This matches ViteHub’s host-independent Workspace/Source primitives and agent-first filesystem affordances with no product-specific surface. The branch is rebased onto `main`; package typecheck, all 164 workspace tests, and all 193 provider tests pass. A one-line test-isolation repair assigns the new base provider-error test a unique chat delivery ID, preserving runtime behavior while preventing Chat SDK deduplication from masking the exercised error path.

onmax:
@codex review

onmax:
@codex review

PR #881, opened 2026-07-31
fix(vite-hub): derive Cloudflare identity from Workers Builds target

onmax:
## Summary

- use Cloudflare Workers Builds' connected Worker as the Cloudflare preset's deployment identity fallback
- reject explicit ViteHub identities that resolve to a different Worker target
- prove the injected identity controls generated Worker, R2, Queue, Rate Limit, Sandbox, Container, and deployment-manifest output

## Why

Cloudflare Workers Builds deploys the connected Worker through `WRANGLER_CI_OVERRIDE_NAME`, but ViteHub previously derived its defaults from `vitehub.name`, `VITEHUB_DEPLOYMENT_NAME`, the package name, or the Vite root. A build could therefore deploy one Worker while generating stateful-resource names for another identity.

The Cloudflare preset now resolves `vitehub.name` → `VITEHUB_DEPLOYMENT_NAME` → `WRANGLER_CI_OVERRIDE_NAME` → package name → Vite root. Matching normalized explicit identities remain valid, while a differing explicit identity fails because Workers Builds keeps the connected Worker as deployment authority. Other presets ignore the Wrangler input.

This derives deterministic resource names; it does not provision the corresponding R2 bucket or Queue. See Cloudflare's [Workers Builds name requirement](https://developers.cloudflare.com/workers/ci-cd/builds/troubleshoot/#workers-name-requirement) and the [Workers SDK name precedence](https://github.com/cloudflare/workers-sdk/blob/5a56ddaf8548fe79787482506b3d5e0233c329c6/packages/workers-utils/src/worker-name.ts).

## Verification

- `pnpm --filter vite-hub test` — 8 test files, 79 tests, and type tests pass
- `pnpm --filter vite-hub typecheck`
- `TMPDIR=/private/tmp pnpm exec vp test test/consumer/deployment-presets.test.ts -t "derives Cloudflare provider output from the Workers Builds target"`
- focused Oxlint and diff checks pass


onmax:
**Verdict: pass**

Head: `5707dd57f17644ab8ea7a6ffd5ed5ff2b651ce9d`

Evidence:
- The implementation directly fixes the stated Workers Builds identity mismatch by using `WRANGLER_CI_OVERRIDE_NAME` only for the Cloudflare preset.
- Explicit ViteHub identities are normalized and accepted when matching, while conflicts with the connected Worker fail rather than generating resources for a different deployment target.
- Unit and consumer coverage exercise Worker, R2, Queue, Rate Limit, Sandbox, Container, and deployment-manifest output; docs state that deterministic naming does not provision R2 or Queue resources.
- The five-file diff contains no unrelated product surface or legacy compatibility work.

onmax:
@codex review

PR #884, opened 2026-07-31
fix(agent): clean up abandoned Harness sessions

onmax:
## Summary

- clean up Harness sessions when eager multi-stream production settles, even if no returned stream is consumed
- route abort cleanup through the runtime lifecycle and share one cleanup task across every result surface
- keep buffered sibling streams readable without imposing a fixed abandonment timeout

## Why

ViteHub creates the Harness session and wraps the returned result, but the existing cleanup wrappers only run after a consumer starts and finishes, returns, throws, or aborts an iterator. A result that is never consumed can therefore leave its session open indefinitely.

Harness exposes producer completion through `responseMessages` and closes its streams before that promise's reactions run. Following that terminal signal lets ViteHub reclaim the session after production finishes while preserving already-buffered output for partial or later consumption.

## Validation

- regression proof on `origin/main`: the three new lifecycle cases failed before the source fix
- `pnpm --filter @vite-hub/agent exec vp test test/invocation-lifecycle.test.ts` — 23 passed
- `pnpm --filter @vite-hub/agent exec vp test --maxWorkers=1` — 68 files and 1,574 tests passed
- `pnpm exec vp check --no-fmt packages/agent/src/harness-agent.ts packages/agent/test/invocation-lifecycle.test.ts`
- `git diff --check`

### DIFF
--- a/packages/agent/src/harness-agent.ts
+++ b/packages/agent/src/harness-agent.ts
@@ -1173,11 +1173,18 @@ async function withSessionCleanup(
   waitUntil: AgentWaitUntil,
 ): Promise<unknown> {
   let cleanupTask: Promise<void> | undefined
+  let markCleanupStarted!: () => void
+  const cleanupStarted = new Promise<void>((resolve) => {
+    markCleanupStarted = resolve
+  })
   let onAbort: (() => void) | undefined
   const cleanupOnce = (error?: unknown) => {
-    cleanupTask ||= cleanup(error).finally(() => {
-      if (onAbort) abortSignal?.removeEventListener("abort", onAbort)
-    })
+    if (!cleanupTask) {
+      cleanupTask = cleanup(error).finally(() => {
+        if (onAbort) abortSignal?.removeEventListener("abort", onAbort)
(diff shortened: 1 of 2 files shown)


onmax:
**Verdict: pass**

Head: `9c3224d653e5d5a938bd1d0a9bd853c3c26ec9f7`

Evidence:
- The intent is a ViteHub runtime-lifecycle concern: an Agent Invocation that is never consumed must not leave its Harness session open indefinitely.
- The change fixes the adapter boundary with the existing `context.runtime.waitUntil` primitive and a shared idempotent cleanup task; it does not expose Harness plumbing in the public API.
- Cleanup follows Harness producer completion or invocation abort while preserving already-buffered sibling streams.
- The exact-head repair makes the producer observer settle through the shared cleanup task after abort, preventing runtime `waitUntil` flushing from hanging.
- Scope remains limited to the adapter implementation and focused lifecycle regression coverage.

onmax:
@codex review

onmax:
@codex review

PR #888, opened 2026-08-01
fix(database): type Cloudflare environment bridge

onmax:
Narrows Cloudflare's generated Worker `Env` interface to ViteHub's record-shaped runtime environment at the generated Nuxt bridge. Without the explicit boundary, consuming Nuxt applications fail typecheck because Cloudflare's concrete interface intentionally has no string index signature.

Runtime behavior is unchanged. The Database suite passes with 12 files and 102 tests, including a regression assertion on the generated middleware.


onmax:
> [!NOTE]
> **Direction verdict: pass**
>
> Head 2f7479ecd7863ed2b830f45be19ecb4c2d1850b4 matches the stated intent: it narrows Cloudflare generated Worker Env only at the generated Nuxt bridge, preserves runtime behavior, and adds a focused assertion for the emitted boundary cast. The two-file change uses the existing bridge primitive and keeps provider-specific typing behind ViteHub framework integration.

onmax:
@codex review

PR #898, opened 2026-08-04
fix(box): remove completed disposable sessions

onmax:
## Summary

- make each validated disposable Crabbox root owner-writable before removing it after normal close or failed bootstrap
- cover the public Box session lifecycle with an exact Git checkout containing read-only nested content

## Why

Read-only reference checkout content can make rm -rf leave a multi-gigabyte /tmp/vitehub-box.* root behind after an agent finishes. Box creates, validates, and owns that disposable root, so consumers should not need package patches to make teardown reliable. The existing exact-root validation and workspace/state safety boundaries remain unchanged; consumer-owned browser session cleanup stays outside this change.

## Validation

- regression proof with the pre-fix cleanup command: public session.close() fails with Permission denied and leaves the disposable root
- pnpm --filter @vite-hub/box exec vp test test/crabbox.test.ts — 26 passed
- tsc --noEmit -p packages/box/tsconfig.json
- vp check --no-fmt packages/box/src/internal/crabbox.ts packages/box/test/crabbox.test.ts
- git diff --check

### DIFF
--- a/packages/box/src/internal/crabbox.ts
+++ b/packages/box/src/internal/crabbox.ts
@@ -287,7 +287,7 @@ function createCrabboxProvider(options: CrabboxSandboxOptions) {
         }
         if (leaseId && remoteRoot) {
           await runCrabbox(sessionOptions, leaseId, {
-            command: `rm -rf -- ${shellQuote(remoteRoot)}`,
+            command: removeDisposableRootCommand(remoteRoot),
           }).catch(() => undefined);
         }
         releaseWorkspace()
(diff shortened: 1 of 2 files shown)


onmax:
**Verdict: proceed**

Head: `932a3f89883d46be3e05754603ac2cb56d2b8060`

Intent: ensure Box-owned disposable Crabbox roots are removed after normal close or failed bootstrap even when a checkout contains read-only nested content.

Evidence:
- The change is confined to `packages/box` cleanup behavior and its regression test.
- Both existing validated cleanup paths now share one command that makes the already validated, Box-owned root owner-writable before deletion.
- The public Box lifecycle test creates an exact Git checkout, adds read-only nested content, closes the session, and asserts the disposable session root no longer exists.
- `origin/main` is an ancestor of this head; no base reconciliation is currently required.
- The change follows ViteHub direction by fixing lifecycle ownership at the Box boundary without adding product-specific surface area.

PR #906, opened 2026-08-05
fix(workspace): avoid redundant source file reads

onmax:
## Summary

- remove root-mounted build source files from metadata already returned by `WorkspaceStore.list()`
- avoid calling `readFile()` once per workspace entry during reconciliation
- prevent GitHub Workspace Store refresh storms and secondary rate limits in Box-backed Codex invocations

## Verification

- `corepack pnpm --filter @vite-hub/workspace test -- source-loader-publisher.test.ts` (54 passed)
- `corepack pnpm --filter @vite-hub/workspace typecheck`
- `git diff --check`

## Live reproduction

Bitácora materialized a root-mounted GitHub workspace with 500+ files. The previous loop called `readFile()` for every entry, and each GitHub store read refreshed branch state, producing a GitHub secondary-rate-limit 403 before Codex could run.

### DIFF
--- a/packages/workspace/src/lifecycle.ts
+++ b/packages/workspace/src/lifecycle.ts
@@ -191,11 +191,9 @@ async function readSyncedBuildSources(store: WorkspaceStore): Promise<SyncedBuil
 
 async function removeRootBuildSourceFiles(store: WorkspaceStore, source: SyncedBuildSource) {
   const entries = await store.list("", { recursive: true })
-  await Promise.all(entries.map(async (entry) => {
-    if (entry.type !== "file") return
-    const file = await store.readFile(entry.path)
-    if (file?.metadata?.source === source.key) await store.rm(entry.path, { force: true })
-  }))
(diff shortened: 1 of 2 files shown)


onmax:
**Verdict: proceed**

Validated exact head `c1050f0040f3711ffba301d87e3e0beaf601d596` against the PR title/body and current `main`.

Evidence:
- The repair removes the per-entry `readFile()` loop and uses metadata already guaranteed by the `WorkspaceStore.list()` entry contract.
- The change stays at the workspace lifecycle boundary and preserves root-source cleanup by deleting only file entries whose `metadata.source` matches the synchronized source key.
- Focused regression coverage proves stale root-mounted source removal without any reconciliation reads.
- The two-file diff is the smallest coherent implementation of the stated GitHub refresh/rate-limit fix; verification and review must still confirm every store preserves file metadata in `list()`.

onmax:
@codex review

PR #914, opened 2026-08-06
fix(agent): type and validate chat routes

onmax:
## Summary

- expose the existing generated chat `route` option through generic `AgentChannelOptions`, so explicit `http({ route })` callers retain admission body and authentication inference
- validate the shared chat envelope after raw-body authentication: optional `id`, `messageId`, and `trigger` strings, a non-empty `messages` array, and public `user` or `assistant` roles
- keep extra request fields intact for additive Standard Schema validation, and document why `http()` and `webChat()` remain distinct Channel Kinds with different route defaults

## Why

`http()` and `webChat()` reach the same generated UI-message route handler, but repository history and current behavior do not support treating them as aliases. `webChat()` deliberately enables the route by default and records the `web-chat` origin, while `http()` remains a generic HTTP Channel whose route is explicit. Those Kinds flow into invocation origin, identity, rate-limit, and webhook behavior, so this preserves the distinction established in [#598](https://github.com/vite-hub/vitehub/pull/598).

The request envelope is common ViteHub protocol rather than application policy. Applications can now remove duplicate loose schemas for the standard AI SDK body and keep admission focused on authentication, trusted context, and any product-specific validation.

## Verification

Passed:

- `pnpm exec vp run -t @vite-hub/agent#build`
- `pnpm exec vp test test/providers.test.ts` (219 tests)
- `pnpm exec vp test test/tutorial-fixture.test.ts`
- full Agent run: 1,734 tests passed before isolating its two unrelated failures; the tutorial fixture passed after building its separate `vite-hub` dependency

`test/box.test.ts` still fails at the existing concurrent colocated-skill installation assertion. The same failure reproduces from the untouched `origin/main` commit `9cc43afb` in a separate worktree, so it is not caused by this diff.

### DIFF
--- a/docs/content/docs/agents/channels.md
+++ b/docs/content/docs/agents/channels.md
@@ -28,6 +28,8 @@ export default defineAgent({
 
 Built-in helpers include `discord()`, `github()`, `http()`, `slack()`, `teams()`, `telegram()`, and `webChat()`. Under a canonical built-in key, pass options directly, such as `channels: { telegram: { messages: { delivery: 'manual' } } }`. A custom Channel id still uses a definition or synchronous factory, so `channels: { portal: webChat }` is equivalent to `channels: { portal: webChat() }`. Use `defineChannel(kind, options)` for an app-owned Channel Kind.
 
+Use `webChat()` for a generated AI SDK chat route; it enables `route` by default and records the `web-chat` Channel Kind. Use `http()` for a generic HTTP Channel; it keeps `route` disabled unless you pass `http({ route: true })` or route options explicitly. Both route-enabled forms use the same request and streaming-response contract, while their distinct Channel Kinds keep invocation origin and delivery semantics honest.
+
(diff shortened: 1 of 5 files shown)


onmax:
**Verdict: proceed**

Validated direction: keep `webChat()` as the default route-enabled web-chat Channel, allow explicit `http({ route })` to use the same generated handler with `http` origin, and validate only the common extensible chat envelope at the ViteHub boundary.

Evidence for head `5f9b48524954bde372c376657c957e07eedd9a96`:
- generated route parsing, authentication, admission, and `chat.message` conversion already share the same ViteHub boundary
- `route` already exists on Channel definitions; this head exposes that existing primitive through generic helper options and preserves distinct defaults/origins
- focused runtime and type tests cover malformed envelopes, raw-body authentication, extension preservation, schema inference, and explicit HTTP routes
- exact-head GitHub checks pass

Lens trace: Evidence—proceed; Precedent—proceed; Synthesis—proceed; Communication—proceed.

onmax:
@codex review

PR #917, opened 2026-08-07
fix(agent): preserve Nuxt chat route context

onmax:
## Summary

- forward the Nitro host event through generated chat handlers into authentication, admission context, and input mapping
- accept readable-stream and UI-message Response shapes across JavaScript realms without relying only on constructor identity
- keep native Web responses on the existing cleanup path, including stale body-header removal and the terminal `[DONE]` frame

## Why

Quiver Portal's Ask AI route crosses the Nuxt/Nitro runtime boundary. The converted Web `Request` preserves HTTP data, but admission also needs the original H3 event for host-owned context. The returned UI message stream can originate in a different runtime realm, where a valid `ReadableStream` or `Response` fails the local `instanceof` check.

This keeps both fixes in the shared Agent route boundary. PR #914 already owns request-envelope typing and validation, so this change leaves that behavior untouched.

## Boundary

The host event remains `unknown` and chat-specific, so H3 types do not leak into ViteHub's host-independent server API. Structural Response recognition requires both a readable body and Headers-compatible `get`; other route families, generated-root placement, and Harness/Codex bundling are unchanged.

## Verification

- focused Agent server-route tests for non-`instanceof` streams and Responses, plus event identity in authentication/context/mapInput
- focused generated Nitro handler assertion
- `corepack pnpm --filter @vite-hub/agent build` (includes publint)
- changed-file `vp lint` passes with only existing warnings outside this diff

A disposable Portal rebuild was not repeated because the shared workspace was under disk pressure; the focused regressions cover the exact downstream patch seams.

### DIFF
--- a/packages/agent/src/server/routes.ts
+++ b/packages/agent/src/server/routes.ts
@@ -105,6 +105,7 @@ export interface AgentTelegramPollingRouteOptions extends AgentRouteRuntimeOptio
 
 export interface AgentChannelChatRouteRequestOptions extends AgentRouteRuntimeOptions {
   agentName?: string
+  event?: unknown
 }
 
 export interface AgentChannelChatRouteStandardSchemaResultSuccess<T = unknown> {
@@ -125,6 +126,7 @@ export interface AgentChannelChatRouteStandardSchemaV1<T = unknown> {
 export interface AgentChannelChatRouteAdmissionContext<TBody extends AgentChannelChatRouteBody = AgentChannelChatRouteBody> {
   agentName: string
   body: TBody
+  event?: unknown
   rawBody: string
   request: Request
 }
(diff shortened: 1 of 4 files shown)


onmax:
**Verdict: revise → correction applied**

Direction: At the generated Agent chat route boundary, forward the opaque host event to admission/mapping callbacks and recognize cross-realm stream/Response values by the capabilities normalization actually uses.

Evidence:
- Head `e65bcb6ea57a81dbc4933cf1988568ecf37c318f` keeps the host event `unknown` and chat-local, preserving the host-independent API boundary.
- Focused tests cover exact event identity plus structural stream, Response, and headers shapes.
- Local predicates now require the capabilities each normalization path invokes; cleanup copies iterable header entries into local `Headers` before removing stale body headers.
- `@vite-hub/agent` typecheck passes; focused provider tests pass 235/235; their prerequisite package build and publint pass.

Required correction: completed. Proceed with exact-head review and merge-gate checks.

onmax:
@codex review

PR #924, opened 2026-08-07
fix(agent): time out Harness attachment resolution

onmax:
Channel attachment callbacks can currently stall forever before a Harness turn starts. This caps each attachment resolution at 15 seconds, while preserving the invocation abort reason when the caller cancels first.

The regression uses a stalled `fetchData()` callback and verifies that the Harness is never invoked and its temporary attachment directory stays clean.

Verification:
- `corepack pnpm --filter @vite-hub/agent exec vp test run test/harness-chat-history.test.ts` (9 passed)
- `corepack pnpm --filter @vite-hub/agent exec vp lint src/harness-agent.ts test/harness-chat-history.test.ts`
- `corepack pnpm --filter @vite-hub/agent build` (including publint)
- Installed `@vite-hub/agent@b0bfc3f` from pkg.pr.new in an empty consumer, then verified a stalled image `fetchData()` rejected after 15134ms without invoking the Harness

### DIFF
--- a/packages/agent/src/harness-agent.ts
+++ b/packages/agent/src/harness-agent.ts
@@ -111,6 +111,7 @@ const harnessGeneratedFiles = ["harness-tool.mjs"] as const
 const harnessInstructionFiles = ["AGENTS.md", "CLAUDE.md"] as const
 const harnessAttachmentDirectory = ".vitehub/attachments"
 const harnessAttachmentMaxBytes = 25 * 1024 * 1024
+const harnessAttachmentResolutionTimeoutMs = 15_000
 const harnessAgentPackage = "@ai-sdk/harness/agent"
 const harnessRemoveDirectory = Symbol.for("vitehub.harnessRemoveDirectory")
 const harnessSandboxAdapter = Symbol.for("vitehub.harnessSandboxAdapter")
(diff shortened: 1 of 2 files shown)


onmax:
**Verdict: proceed**

**Exact head:** `b0bfc3f0a1eed10ae6a2b034254c1385bdf7e189`

**Intent:** bound each Harness attachment callback at 15 seconds while preserving caller-abort reasons and cleaning pre-invocation attachment state.

**Evidence:**
- Harness materialization awaits Channel `fetchData()` before invoking the Harness; the shared callback contract otherwise permits a promise that never settles.
- The deadline stays private at the Harness-owned attachment boundary, alongside its size and sandbox-cleanup policy, rather than imposing this lifetime on all message consumers.
- Timer and abort-listener cleanup is shared across resolution, rejection, timeout, and cancellation; caller cancellation retains its original abort reason when it wins.
- Focused coverage uses a stalled callback and verifies timeout before any Harness turn with no temporary attachment path left behind; the adjacent abort regression covers cancellation precedence.
- The only material residual risk is the unmeasured 15-second value, which is a narrow, reversible private constant and does not outweigh the demonstrated indefinite stall.

**Direction result:** proceed with base reconciliation, verification, and exact-head review.

onmax:
@codex review

PR #931, opened 2026-08-07
fix(agent): use local state for Vite serve

onmax:
When automatic Agent state runs under a Cloudflare production preset, Vite development currently inherits the deployment provider and installs Cloudflare Durable Object state. That prevents the local runtime from using the SQLite default it already resolves for `command: serve`.

This passes Vite’s command into the early config hook and makes automatic Cloudflare state selection yield to local development. Explicit Cloudflare state providers and the `cloudflare-agents` runtime keep their existing behavior.

Verified with:

- all 237 Agent provider tests
- Agent typecheck
- Agent package build and publint
- a disposable Calories checkout with the Agent patch removed and this package installed: Nuxt dev started under `vitehub.preset: cloudflare`, `/` returned HTTP 200, and the generated handler used `.data/vitehub-agent-state.sqlite` without a Cloudflare state import

### DIFF
--- a/packages/agent/src/vite.ts
+++ b/packages/agent/src/vite.ts
@@ -482,6 +482,7 @@ function resolveAgentHosting(config: unknown): "cloudflare" | "netlify" | "verce
 function shouldInstallCloudflareAgentState(
   options: false | ResolvedAgentModuleOptions,
   config: unknown,
+  command?: "build" | "serve",
 ): options is ResolvedAgentModuleOptions {
   if (!options) return false
   const { provider, url } = options.providers.state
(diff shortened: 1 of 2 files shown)


onmax:
**Verdict: proceed**

The PR fixes a real lifecycle mismatch: later Agent state resolution already treats Vite `command: serve` as local development, while the early config hook only saw the Cloudflare production preset and installed Cloudflare state. The head passes the hook environment command into the existing provider predicate, preserves the earlier precedence of explicit Cloudflare providers and `cloudflare-agents`, and adds focused coverage that asserts no Cloudflare output plus generated local SQLite state.

Required follow-through: reconcile with current `main`, rerun verification, and obtain positive exact-head review evidence before merge.

Validated head: `a86955a9fab2521583e69d606608479a9ecf5960`
Intent: `fix(agent): use local state for Vite serve`

onmax:
@codex review

PR #933, opened 2026-08-08
fix(agent): preserve Nitro 3 webhook queues during shutdown

onmax:
## Summary

- Stop generated libSQL webhook queue workers when an existing Node framework owner handles `SIGINT` or `SIGTERM`.
- Prepend ViteHub's queue cleanup so its promise reaches the request-derived `waitUntil` set before srvx begins closing.
- Keep the Nitro close hook for Nitro 2, and preserve Node's default signal termination when no framework shutdown listener exists.

## Why

The pinned Nitro 3 Node entry starts srvx but never calls the Nitro `close` hook. srvx owns signal handling and waits for request `waitUntil` work, but it cannot infer that ViteHub must call the generated queue worker's `stop()`. An active delivery can therefore keep srvx's wait set pending until forced shutdown instead of being aborted and returned to durable storage.

The bridge is installed only after queue ownership starts and only for signals that already have a framework listener. It removes only its own listeners, never calls `process.exit` or closes the server, and reuses the same captured request lifetime that queue work already uses.

Motivating lifecycle review: https://github.com/PostHog/posthog-js/pull/4445#discussion_r3738879168

## Verification

- Reproduced before the fix with a packed Agent consumer on `nitro@3.0.260603-beta`, `h3@2.0.1-rc.25`, and the real Node preset: an active queued delivery received SIGTERM, but srvx remained in its graceful-shutdown countdown because the queue task never settled.
- The same packed-runtime test now proves:
  - no framework listener under `CI` leaves SIGTERM to Node's default `[null, "SIGTERM"]` exit;
  - the production listener order is `shutdownWebhookQueues` then srvx `shutdown`;
  - duplicate SIGTERM does not duplicate or race cleanup;
  - the active invocation observes abort, delayed cleanup completes, and the child exits with code 0;
  - the durable delivery is immediately claimable again from SQLite.
- Focused generated-provider test passes.
- `@vite-hub/agent` typecheck passes.
- `@vite-hub/agent` build and publint pass.
- Changed-file lint reports 0 errors and 6 existing warnings in `providers.test.ts`.
- The full Agent suite reached 1,778 passing tests and 2 unrelated deterministic setup failures: the Box concurrency test's sandbox mock rejects colocated-skill installation, and the tutorial fixture lacks prebuilt workspace package outputs.


onmax:
**Verdict: proceed**

Head `8ceb395968029de4dea7992ac1268d813abcb4c6` matches the stated intent.

- The change stays at the generated Agent/Nitro integration boundary that owns webhook queue startup and shutdown.
- It uses existing queue `stop()` and request-derived `waitUntil` primitives; it does not add a product surface or framework-owned exit policy.
- Signal bridging is limited to Node signals already owned by a framework listener, while Nitro 2 retains the existing `close` hook path.
- The focused packed Nitro 3 lifecycle test covers default Node termination, listener ordering, idempotent cleanup, abort completion, clean exit, and durable re-claim.

onmax:
@codex review

PR #939, opened 2026-08-08
feat(email): add Resend delivery adapter

onmax:
## Summary

- add a dependency-free Resend adapter at `@vite-hub/email/drivers/resend` and `vite-hub/email/drivers/resend`
- resolve API keys on every send so request-scoped Worker secrets remain safe
- map the portable ViteHub email contract, including attachments, and normalize Resend failures into existing `EMAIL_*` errors

## Why

ViteHub already owns the portable `EmailDriver` contract, but consumers still need to reimplement Resend payload and error plumbing. This adds the missing provider seam while keeping provider selection in application server code and credentials out of Vite config.

Resend has no send-side equivalent for `contentDisposition`, so the adapter rejects that field explicitly instead of silently dropping it.

## Verification

- `pnpm --filter @vite-hub/email test` — 53 tests passed
- `pnpm --filter vite-hub test` — 98 tests passed
- `pnpm --filter @vite-hub/email typecheck`
- `pnpm --filter vite-hub typecheck`
- packed framework published-types fixture passed with installers disabled
- Formula 100 throwaway `pnpm patch` proof passed its email tests and typecheck using the public adapter with a runtime secret getter
- independent code review found no remaining actionable issues

A later narrow package-contract rerun hit the existing 10-second per-test timeout, so the packed fixture was also run directly and passed. No live Resend API send was performed as part of this PR.


onmax:
**Verdict: proceed**

Head `ea479b216920d83edd94acda27ebedfe53e8103a` matches the PR intent: it adds a dependency-free Resend implementation behind the portable `EmailDriver` boundary, exposes and documents both package surfaces, resolves credentials per send, quotes structured addresses safely, and bounds stalled provider requests with a normalized `EMAIL_TIMEOUT`. Package tests, framework tests, contract tests, and both typechecks pass locally.

onmax:
@codex review

onmax:
@codex review

PR #963, opened 2026-08-12
fix: retire Calories ViteHub patches

onmax:
## Summary

- activate discovered Cloudflare Agent Workflows only when a required durable channel handoff has its exact generated binding
- keep channel timeouts at the webhook and typing boundary, including prompt cleanup when a durable handoff fails
- preserve colocated Agent instructions in generated Workflow registries

## Patch audit

Provider cost, manual delivery defaults, Workflow portability, database capabilities, retry translation, and Cloudflare externalization are already upstream. Reply context remains owned by #948, while the patched global `__env__` mutation is obsolete because Cloudflare environments are scoped with `AsyncLocalStorage`.

## Verification

- `pnpm exec vp run build`
- `pnpm --filter @vite-hub/agent typecheck`
- `pnpm --filter @vite-hub/workflow typecheck`
- `pnpm --filter @vite-hub/agent exec vp test test/providers.test.ts`
- `pnpm --filter @vite-hub/agent exec vp test test/test-runner.test.ts`
- `pnpm --filter @vite-hub/workflow exec vp test`
- `pnpm lint`
- isolated Calories install, focused channel tests, typecheck, and production build against a composite pkg.pr.new preview of this PR plus the current #948 and #961 heads, with only the two `@vite-hub/*` patches removed

PR #968, opened 2026-08-13
fix: harden hosted consumer runtime boundaries

onmax:
Repair three independent ViteHub-owned runtime boundaries exposed by one hosted-consumer validation path; this does not introduce a shared hosted-consumer abstraction.

A production Telegram delivery failure disappeared because Channel effect errors were only attached to optional traces. The shared delivery boundary now emits one best-effort structured console record with capped dynamic fields for the channel, effect, intent, run, and provider error while preserving existing invocation and no-retry behavior.

Installing that fix into the Calories consumer exposed two current-main output regressions. Generated Cloudflare Blob middleware required a consumer-owned ambient declaration, and hosted Database aliases followed Nuxt `srcDir` or a custom Vite `root` even though ViteHub generated both runtime modules under its project root. Generated Blob middleware now handles the provider virtual-module import without a consumer shim, and Database aliases resolve from the ViteHub project root that owns the generated files.

## Verification

- Agent runtime suite: 366 tests passed
- Agent package typecheck
- Blob generated-middleware typecheck with and without consumer Cloudflare declarations, plus package typecheck
- Database Nuxt suite: 24 tests passed
- Database package typecheck
- Aggregate ViteHub Nuxt integration expectation
- `git diff --check`

### DIFF
--- a/packages/agent/src/index.ts
+++ b/packages/agent/src/index.ts
@@ -1000,6 +1000,15 @@ async function applyChannelDeliveryEffectIntents<
       }
       catch (error) {
         delivered = false
+        console.error(JSON.stringify({
+          scope: "vitehub.channel.delivery",
+          event: "outbound.failed",
+          channelId: active.channelId,
+          effect: intent.kind,
+          intent: intent.intent,
+          runId: context.run?.runId,
+          error: agentErrorMessage(error).slice(0, 2_000),
+        }))
         await traceAgentChannelDeliveryEffect(toTraceContext(context), intent, {
           ...metadata,
           "error.message": agentErrorMessage(error),
--- a/packages/agent/src/index.ts
+++ b/packages/agent/src/index.ts
@@ -1000,15 +1000,18 @@ async function applyChannelDeliveryEffectIntents<
       }
       catch (error) {
         delivered = false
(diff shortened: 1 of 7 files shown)


onmax:
**Verdict: `revise` (applied)**

Validated exact local head `032dcf0c7367ec95593dd906e948d574e8129bc5` against the revised PR intent and current `main` (`e3d4404c66e203f90c0ebb174165a3ebf473df34`). Each fix is localized at its owning ViteHub boundary and absent from `main`; PR #963 and the documented Calories validation path support the bounded grouping. The strongest objection was that three independent contracts could be mistaken for a shared hosted-consumer abstraction.

Correction applied: the title/body now state the validation relationship and reject a shared abstraction. The rebased exact head independently passes the Agent regression and typecheck, Blob suite/typecheck, Database suite/typecheck, aggregate ViteHub suite/typecheck, and `git diff --check`. Two-axis review found no spec defects or hard standards violations.

Evidence artifacts: `/tmp/validate-direction/vitehub/pr-968/brief.md` and `/tmp/validate-direction/vitehub/pr-968/verdict.md`.

onmax:
@codex review

PR #975, opened 2026-08-14
fix(schedule): discover Nuxt server suffix definitions

onmax:
Nuxt can use `app/` as its Vite root while keeping Nitro definitions in the sibling `server/` directory. Schedule suffix discovery only scanned the Vite root, so `server/monthly.schedule.ts` was silently omitted and no provider cron was generated.

This includes resolved server directories when they sit outside the existing Vite scan roots, while excluding `server/schedules/` from suffix discovery so the directory convention remains owned by its existing source.

## Verification

- `pnpm --filter @vite-hub/schedule test` — 198 tests passed
- Packed Nuxt Cloudflare consumer imports `server/monthly.schedule.ts` in the generated registry, emits `0 8 1 * *` in the final Wrangler config, includes the `cloudflare:scheduled` hook, and excludes the MCP/PKCE dependency closure.

### DIFF
--- a/packages/schedule/src/discovery.ts
+++ b/packages/schedule/src/discovery.ts
@@ -1,5 +1,5 @@
 import { readFileSync } from "node:fs"
-import { resolve } from "node:path"
+import { isAbsolute, relative, resolve, sep } from "node:path"
 
 import {
   createDirectoryDefinitionSource,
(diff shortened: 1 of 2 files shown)


onmax:
> [!NOTE]
>
> Head `d7519b09fe5847e722bc6fce2567f0994e6879ba` matches the PR intent. The change extends suffix discovery only to resolved server definition directories outside existing Vite scan roots, filters each server directory's `schedules/` subtree so the established directory source remains authoritative, and adds regression coverage for both sibling suffix and directory conventions. Current `main` does not already contain the fix, and the intervening base change is unrelated.

onmax:
@codex review

PR #993, opened 2026-08-19
feat(agent): improve channel message handling

onmax:
Channel consumers currently need app-level plumbing to classify provider failures and recover typed structured output, while correction-heavy chats cannot express steering intent. This moves those shared seams into the Agent boundary.

- Normalize AI SDK authentication, quota, rate-limit, and availability failures into sanitized public errors, expose `publicError` to hooks, and use those errors in the default channel fallback.
- Infer Standard Schema output in finish hooks so consumers can use `event.result` without a parallel type declaration.
- Accept `messages.concurrency: "steer"`; Chat uses queue semantics until model and Workflow runtimes expose the same resumable-input contract, which is recorded as an implementation TODO.

## Verification

- `corepack pnpm --filter @vite-hub/agent build`
- Provider public-error tests: 4 passed
- Chat error/concurrency tests: 7 passed
- Durable concurrency test: 1 passed

The full package suite was also attempted. After fixing a test update-ID collision introduced here, eight unrelated tests remained blocked by fresh-worktree timeouts or missing root build output; all changed-path tests pass independently.

### DIFF
--- a/packages/agent/src/agent-error.ts
+++ b/packages/agent/src/agent-error.ts
@@ -96,6 +96,10 @@ export type AgentPublicErrorCode =
   | "CAPABILITY_NOT_FOUND"
   | "INTERNAL"
   | "LLM_GATE_REJECTED"
+  | "PROVIDER_AUTHENTICATION_FAILED"
+  | "PROVIDER_QUOTA_EXHAUSTED"
+  | "PROVIDER_RATE_LIMITED"
+  | "PROVIDER_UNAVAILABLE"
   | "RATE_LIMIT_REJECTED"
   | "RATE_LIMIT_UNAVAILABLE"
   | "TRANSCRIPTION_AUTHENTICATION_FAILED"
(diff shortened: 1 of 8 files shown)


onmax:
**Verdict: proceed**

**Head:** `7ffd66c63447f340aa28dcb59f1825fa6f2e1d0c`

**Intent:** Move shared channel-message error classification and structured-output typing into the Agent boundary, and accept steering intent without changing Chat's current queue semantics.

**Evidence:**
- Provider failures are normalized through the existing `AgentPublicError` boundary and exposed to both Agent and Chat error hooks without leaking provider payloads.
- Standard Schema output propagates through the Agent definition into finish-hook `event.result` typing.
- `messages.concurrency: "steer"` is accepted and deliberately maps to Chat queue behavior pending a shared resumable-input contract.
- The head adds focused runtime, provider, public-error, and type tests; the exact-head repository CI, docs, consumer, Cloudflare, Vercel, and Netlify checks pass.
- The source changes are the deliberate replay of the feature commit after its temporary base-branch revert, leaving a coherent eight-file diff against current `main`.

_Note: the named `validate-direction` skill was unavailable in the session catalog and filesystem, so this verdict was reproduced from its required exact-head/intent/evidence contract._

PR #1004, opened 2026-08-21
fix(agent): decode workflow audio data URLs before transcription

onmax:
Workflow portability serializes binary audio attachments as data URLs, but Agent transcription forwarded those strings unchanged to AI SDK 7. AI SDK treats string audio as raw base64 and rejects the full data URL before calling the transcription provider.

Decode only `data:` audio strings at the shared transcription boundary. Raw base64 strings keep their existing behavior, and `maxBytes` applies to the decoded bytes.

## Verification

- `corepack pnpm exec vp test run test/transcription.test.ts` (36 passed)
- `node --test tests/openrouter-transcription.test.mjs` in Calories (downstream patched-consumer proof)
- `git diff --check`

The full Agent suite is not claimed as passing. It hit widespread unrelated 5-second timeouts across other test files on this machine and was stopped; the focused transcription suite stayed green.

### DIFF
--- a/packages/agent/src/capabilities/transcribe.ts
+++ b/packages/agent/src/capabilities/transcribe.ts
@@ -1,5 +1,5 @@
 import { defineCapability } from "../capability-runtime.ts"
-import { appendMessageText } from "../messages.ts"
+import { appendMessageText, attachmentStringByteLength, attachmentStringBytes } from "../messages.ts"
 import { loadAiSdk } from "../internal/ai-sdk-runtime.ts"
 import { isTranscriptionAbortError, isTranscriptionError, transcriptionError } from "./transcription.ts"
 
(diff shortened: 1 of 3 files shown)


onmax:
Verdict: proceed
Direction: Workflow-portable audio attachments must reach model-backed Agent transcription as decoded bytes; normalize only data: strings at the shared transcription-to-AI-SDK boundary while raw base64 and other audio forms remain unchanged.
Evidence: Exact-head 3f8e5b346c73943b3680f31fb1e8fe2065c76152 changes the Agent transcription boundary plus the shared attachment decoder's byte-length preflight. Focused coverage proves direct and lazy base64 data URLs, percent-encoded binary octets, raw-base64 preservation, decoded-size enforcement before allocation, and no provider call for oversized bytes.
Philosophy: The shared attachment decoder owns data URL byte semantics, while transcription owns its maxBytes policy. The branch composes those existing boundaries so Workflow audio reaches AI SDK correctly without duplicating decoding or weakening resource limits.
Challenge: Data URLs must neither bypass maxBytes nor be rejected by encoded length, and percent-encoded binary must not pass through UTF-8 URI decoding. The exact head computes decoded byte length without allocating the payload, enforces the limit, then uses attachmentStringBytes(); raw base64 retains its prior AI SDK behavior.
Next: Await exact-head checks and Codex review, verify all threads and child pull requests, then merge only if the gate remains clean.

onmax:
@codex review

onmax:
@codex review

onmax:
@codex review

PR #1006, opened 2026-08-21
fix(workspace): avoid per-file executable probes

onmax:
A Workspace Session currently shells out to `test -x` once for every host file. Babysitter provider workspaces can contain about 130,000 dependency files, which leaves teardown running for nearly an hour after the Agent Invocation timeout.

This change lets a Workspace host report the executable bit in its file listing. The local host already calls `lstat()`, so it now returns that bit with the file size. Hosts that do not report it keep the bounded `test -x` fallback.

Verification:
- Workspace host-session suite: 49 passed
- Agent provider suite: 52 passed
- Workspace and Agent typechecks passed
- Regression loop drops executable probes from 4 to 0 while preserving `100755`

### DIFF
--- a/packages/agent/src/provider-agent.ts
+++ b/packages/agent/src/provider-agent.ts
@@ -385,7 +385,12 @@ export function localWorkspaceHost(): WorkspaceSessionHost {
           for (const entry of await readdir(directory, { withFileTypes: true })) {
             const target = join(directory, entry.name)
             const type = entry.isSymbolicLink() ? "symlink" : entry.isDirectory() ? "directory" : "file"
-            entries.push({ path: target, ...(type === "file" ? { size: (await lstat(target)).size } : {}), type })
+            const stats = type === "file" ? await lstat(target) : undefined
+            entries.push({
+              path: target,
+              ...(stats ? { executable: Boolean(stats.mode & 0o100), size: stats.size } : {}),
+              type,
+            })
             if (options?.recursive && type === "directory") await visit(target)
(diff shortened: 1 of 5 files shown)


onmax:
@codex review

onmax:
**Verdict: revise**

Head: `76571ac8f28fe7b507e7c13ef10c1c773e63ef5e`

Intent: avoid one `test -x` subprocess per listed host file while preserving Workspace Git-mode behavior.

Decisive evidence: the local host already has `lstat()` data, so reporting an executable fact at that boundary removes the probes. Exact-head review identified two representation bugs in the prepared commit: `stats.mode & 0o111` did not match owner execution semantics, and mapping a known false result to public `100644` metadata broke the existing canonical form.

Required correction applied: local executability now uses `0o100`; a host-reported false value suppresses fallback probing without becoming metadata; only executable files and symlinks expose Git-mode metadata.

- Evidence: focused tests cover host-reported true, false, omitted values, and executable-to-regular transitions; both package typechecks pass.
- Precedent: GitHub-backed reads normalize `100644` to absent metadata, while executable files use `100755`.
- Synthesis: the host listing owns the executable fact; Workspace conversion owns canonical public metadata.
- Communication: the PR remains scoped to probe elimination and its regression coverage records each reviewed invariant.

Next action: exact-head Codex review and merge-gate verification.

onmax:
@codex review

onmax:
@codex review

PR #1019, opened 2026-08-22
refactor: remove prelaunch migration paths

onmax:
ViteHub is prelaunch, so publishing an application migration path and carrying removed-option shims implies an upgrade contract the project does not have. This removes that contract and leaves current APIs as the only supported shape.

## Current contract

- Removes the migration page, skill route, and cross-links.
- Deletes upgrade-only option guards, generated-file cleanup, and discovery exclusions.
- Makes Sandbox package entries function-only and `inputCommands().call` the single handler name.
- Removes `VITEHUB_DEPLOYMENT_NAME`; explicit `vitehub.name`, Workers Builds identity, package name, and root fallback remain.
- Keeps active standards and provider behavior, including Agent Invocation identity propagation, database/Durable Object migrations, and the Queue name used by the current Drop deployment.

## Verification

- Affected package and docs typechecks pass.
- Docs: 54 tests pass.
- Env: 13 tests pass.
- Sandbox: 232 tests pass.
- ViteHub: 71 tests pass.
- Workspace focused concurrency suite: 49 tests pass.
- Agent changed suites: 973 tests pass; two unrelated provider tests hit their existing local timeouts under the combined run, while the changed invocation-control cases pass in isolation.
- `fallow:dead-code`, `knip:catalog`, lint, and `git diff --check` pass.
- Six of seven deployment-preset tests pass; the Deno execution case cannot run locally because the `deno` binary is unavailable.

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>


onmax:
**Verdict: revise, applied**

The cleanup direction matches the prelaunch contract, but the original exact-head CI showed that one removed path was current runtime behavior. `packages/agent/src/provider-agent.ts` still passes provider control identity through `context.runtime.run.runId`; deleting that lookup caused six failures across live approval, provider input, and queued invocation tests. The branch now keeps the `runId` lookup and its focused regression assertion while retaining the remaining migration-only deletions.

- Evidence: verified head `0925fff93145f101d1dda3d28377eb43f35fb11e`, base `b2fb92beefacede323a56df1f39690adf5156273`; provider failure set 339/339, Markdown-template 29/29, tutorial fixture 2/2, Agent typecheck, lint, and diff checks pass.
- Precedent: prelaunch public shims may be removed, but commit `97e4b3fe` introduced the lookup for current webhook steering.
- Synthesis: preserve internal invocation identity while deleting obsolete public upgrade contracts.
- Communication: the PR body states that current Agent Invocation identity propagation remains supported, and the Deno deployment docs distinguish required organization identity from the optional app override.

Next action: await exact-head checks and review, then merge when the gate holds.

onmax:
@codex review

onmax:
@codex review

PR #1024, opened 2026-08-24
test(agent): restore steer lease timing margin

onmax:
Restore the ambiguous steered Workflow fixture lease from 40ms to 400ms while preserving the exact two-lease overlap and `createBatch === 3` assertions.

Main CI demonstrated both timing directions at 40ms: run 32692295717 observed five calls, while run 32692950394 observed only two.

Focused command (no build):
`../../node_modules/.bin/vp test run test/providers.test.ts -t "keeps an ambiguously accepted steered Workflow input claimable across restarts" --pool=threads --maxWorkers=1`

Fresh-clone result: the test still reached the earlier three-call wait with only two calls at 400ms. This PR intentionally contains only the requested timing-margin restoration so the remaining ownership timing can be evaluated without mixing source changes.

### DIFF
--- a/packages/agent/test/providers.test.ts
+++ b/packages/agent/test/providers.test.ts
@@ -14559,10 +14559,12 @@ describe("server helpers", () => {
 
       await vi.waitFor(() => expect(createBatch).toHaveBeenCalledTimes(3), { timeout: binding!.steer!.ttlMs * 5 })
       await new Promise((resolve) => setTimeout(resolve, binding!.steer!.ttlMs * 2))
+      const acquireLock = vi.spyOn(state, "acquireLock")
       const overlappingDelivery = handler(chatWebhookRequest(91_145), "telegram", {
         agentIdentity: { name: "calories" },
         cloudflare: { env },
       })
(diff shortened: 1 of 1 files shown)


onmax:
> [!NOTE]
> **Direction verdict: revise**
>
> Keep the 400ms synthetic lease and the exact createBatch === 3 and two-lease overlap assertions. Scale the three-call wait timeout to that lease so the fixture allows the ownership handoff it asserts. Do not change runtime ownership behavior in this PR.
>
> Evidence: exact-head CI failed only at the three-call wait after 1,473ms, observing two calls. The wait uses Vitest default timeout even though the fixture now needs a 400ms lease expiry and recovery before the third call.
>
> Evidence lens: the earlier fresh-clone result and exact-head CI reproduce the same timeout. The strongest uncertainty is host scheduling, so the wait must derive its budget from the fixture lease.
>
> Precedent lens: nearby tests preserve explicit ownership assertions and give asynchronous handoffs bounded waits.
>
> Synthesis lens: change only the wait budget to ttlMs * 5. This retains the restored margin and all ownership assertions without changing production behavior.
>
> Communication lens: describe the repair as lease-aware test synchronization, not an ownership-runtime fix.

onmax:
@codex review

onmax:
@codex review

onmax:
@codex review

PR #1025, opened 2026-08-24
fix: bound provider and schedule shutdown

onmax:
A Provider turn can finish successfully while runtime, Workspace, or temporary-root cleanup remains stuck. That left the caller waiting indefinitely, retained multi-gigabyte provider roots in the main Node process, and made cleanup-only timeouts fatal. Separately, generated Process Schedule runtimes admitted queued work after SIGTERM because they did not close until Nitro's later shutdown hook.

This change makes the owning lifecycle boundaries bounded and cancellable:

- hosted Workspace traversal and `diff`/`commit`/`close` propagate an AbortSignal;
- Provider cleanup receives a separate 10-second post-success window, reports cleanup-only timeout as recoverable, and removes guarded `/tmp/vitehub-provider-*` roots in an external process;
- generated Nitro schedule wiring closes the Process Schedule runtime immediately on SIGINT/SIGTERM and reuses the same idempotent close path for Nitro shutdown.

Genuine invocation failures remain fatal, late provider startup retains its Workspace until provider shutdown completes, and independent session locks are released only at the bounded cleanup boundary.

## Verification

- `packages/workspace`: `pnpm exec tsc --noEmit -p tsconfig.json`; `pnpm exec vp test run test/host-session.test.ts` (52/52)
- `packages/schedule`: `pnpm exec tsc --noEmit -p tsconfig.json`; `pnpm exec vp test run test/vite.test.ts` (33/33)
- `packages/agent`: `pnpm exec tsc --noEmit -p tsconfig.json`; `pnpm exec vp test run test/provider-agent.test.ts` (67/67)
- `git diff --check origin/main...HEAD`

No build was run.


onmax:
@codex review

onmax:
**Verdict: revise**

Keep the bounded cleanup signal for host inspection, restoration, and the wait to enter Workspace publication. Once durable publication starts, finish the full diff and snapshot without another cancellation check. Apply the same rule to pre-publication unsafe-symlink replacement: cancellation may stop before removal, but once removal begins, finish writing the inert replacement without the signal.

Keep Process Schedule close tied to the host shutdown lifecycle. ViteHub must start its idempotent close on SIGINT/SIGTERM and let Nitro enforce the process deadline; a second ViteHub deadline would race or preempt the host grace period and is outside this PR contract.

- Evidence: `commitHostChanges` mutates entries one at a time, and unsafe-symlink sanitization removes the host path before rewriting it. Cancellation between those operations can leave the path missing. Nitro owns server termination and its graceful deadline.
- Precedent: the publication lock and snapshot define the durable Workspace boundary. Destructive replacement is likewise one indivisible mutation. Generated integrations cooperate with their host lifecycle instead of replacing its termination policy.
- Synthesis: check cancellation before publication and before each replacement, but do not interrupt either durable writes or a replacement after its first mutation. Start Schedule close on the first signal without adding a competing process deadline.
- Communication: a cleanup timeout may stop waiting for publication or the next replacement, but it must not cause a partial commit or deleted host path. A stalled Schedule close remains subject to the host deadline.

The repair removes the signal from the replacement write and covers cancellation after removal while the inert file write and publication still finish.

onmax:
@codex review

PR #1030, opened 2026-08-24
perf(workspace): avoid redundant preparation work

onmax:
## Summary

- checkpoint lazy Source metadata at lifecycle boundaries instead of rewriting the full manifest after every item
- avoid hashing every Local Workspace entry during ordinary `list()` calls while preserving digest-bearing snapshots and publications
- add an explicit `writeBack: false` hosted Session mode and infer it for read-only Agent Workspace facades
- restore non-publishing Sessions from authoritative Workspace entries without a host-wide baseline scan
- skip unused close-time snapshots and bound digest hydration concurrency

## Production-shaped benchmark

Synthetic Quiver workload: 3,033 files / 10.3 MB.

| Phase | Before | After (3 runs) | Improvement |
| --- | ---: | ---: | ---: |
| Cold preparation | 28.949 s | 6.453–7.535 s | 74.0–77.7% faster |
| Warm preparation | 16.818 s | 1.922–2.871 s | 82.9–88.6% faster |
| Session close | 16.204 s | 2.206–2.390 s | 85.3–86.4% faster |

The original Source loop performed 3,002 metadata writes and serialized about 930 MB for a final ~624 KB manifest. Boundary checkpoints reduce that to two writes.

## Verification

- 106 affected Workspace tests passed, no type errors
- 68 provider Agent tests passed
- strict TypeScript Doctor: 100/100, no findings
- `@vite-hub/workspace` and `@vite-hub/agent` builds passed
- downstream Quiver lint, typecheck, production build, and repeated `<10 s` benchmark budget passed


onmax:
@codex review

PR #1062, opened 2026-08-26
feat(agent): package opt-in Codex runtime

onmax:
## Summary

- resolve an app-installed `@openai/codex` CLI through T3's `binaryPath`
- copy only the current macOS/Linux native payload into production self-hosted Node output
- preserve the host `codex` command fallback when the package is absent or the host is Windows
- reuse the existing Node runtime package copier and document the same-host build contract

## Before / after

| Before | After |
| --- | --- |
| Apps maintain a private deployment workspace, run `pnpm deploy`, copy Codex into the image, and rewrite `PATH`. | Apps run `pnpm add @openai/codex@0.149.1`; ViteHub resolves it and packages the matching payload for production Node builds. |
| Generic dependency tracing copies every Codex platform package (~1.8 GB in the research fixture). | ViteHub selects one native package (~309 MB for Linux x64 in the real-package smoke test). |
| Auth and executable setup are mixed in app Docker plumbing. | ViteHub owns executable packaging; the deployment environment still owns credentials and secrets. |
| Non-Codex apps risk inheriting a large binary dependency. | ViteHub does not declare or install Codex; non-Codex apps remain unchanged. |

Production packaging is deliberately limited to macOS/Linux builds targeting the same OS and CPU architecture. Dev/watch and non-Node Nitro targets do not copy the runtime.

## Verification

- `@vite-hub/agent`: 77 focused resolver/provider tests passed
- Nitro packaging contract: 1 passed, including production, dev, and non-Node behavior
- `@vite-hub/internal`: 40 deployment-output tests passed; copied output executes successfully
- agent and internal typechecks passed
- agent and internal package builds passed, including publint
- changed-file lint: 0 errors; 5 unrelated existing warnings
- real Linux x64 package smoke test: copied wrapper returned `codex-cli 0.149.1`, with only `@openai/codex` and `@openai/codex-linux-x64` present

Downstream Quiver validation should build its Nitro image from this package and run an authenticated `codex app-server` invocation before deleting the existing `runtime/codex` workspace.

### DIFF
--- a/docs/content/docs/agents/agent-drivers.md
+++ b/docs/content/docs/agents/agent-drivers.md
@@ -68,6 +68,12 @@ The `model` value may also be a compatible AI SDK model or an invocation-time ca
 
 The built-in Drivers reuse T3 Code's normalized Codex and Claude Code runtime while ViteHub owns Agent Definitions, Capabilities, Workspaces, Invocations, and public lifecycle events.
 
+Install the Codex CLI as a project dependency when you select the Codex Driver:
+
+```sh
+pnpm add @openai/codex@0.149.1
+```
+
 ```ts [server/agents/review/agent.ts]
 import { defineAgent } from 'vite-hub/agent'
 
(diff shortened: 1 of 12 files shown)


onmax:
continue

PR #1063, opened 2026-08-26
feat(agent): package opt-in Claude runtime

onmax:
## What

Extends the opt-in Codex packaging from #1062 to the built-in Claude Code Driver.

- resolves an app-installed `@anthropic-ai/claude-agent-sdk` and passes its native `claude` executable to T3
- copies the SDK plus exactly one OS/CPU/libc native package into production Node output
- preserves the host-command fallback when the native package is unavailable at runtime
- documents the explicit provider installs and same-host build contract

## Before / after

| Before | After |
| --- | --- |
| Apps needed deployment-owned Claude copy/setup logic. | `pnpm add @anthropic-ai/claude-agent-sdk` is enough for ViteHub to resolve and package Claude. |
| The provider runtime installed the Claude SDK for Codex-only consumers. | The SDK is loaded and installed only by apps that select Claude. |
| Linux native selection ignored libc. | Node builds select the matching glibc or musl package. |

## Dependencies

#1062 is merged. This PR pins the provider-runtime preview from vite-hub/t3code#2, which removes the SDK dependency and defers its import until Claude starts. The catalog can move to the merged T3 artifact before this PR lands.

## Verification

- 398 focused provider/runtime/Nitro tests passed before the branch refresh
- focused provider runtime package tests: 6 passed after the branch refresh
- strict TypeScript doctor passed after the branch refresh
- `@vite-hub/agent` typecheck and package build passed before the branch refresh
- changed-file lint passed with five pre-existing warnings before the branch refresh
- a clean pnpm consumer imported `@anthropic-ai/claude-agent-sdk@0.3.246`; the resolver selected its real Linux x64 glibc package and executed Claude Code `2.1.246`
- `git diff --check` passed after the branch refresh


onmax:
@codex review

PR #1105, opened 2026-08-27
fix(box): isolate Crabbox static sessions

onmax:
Crabbox static-host sessions now receive a unique, stable `CRABBOX_STATIC_ID` for each ViteHub Box session. Concurrent Agent Invocations no longer compete for Crabbox's implicit shared static identity, while non-static Crabbox profiles keep their existing environment behavior.

## Verification

- New concurrent static-host regression test passed.
- `@vite-hub/box` typecheck passed.
- Package build and publint passed.
- The full Crabbox file passed 27 of 28 tests; the unrelated Linux process-reparenting timing test failed because the spawned fixture remained parented to the test runner instead of PID 1.


onmax:

PR #1148, opened 2026-08-27
feat(agent): add adaptive process capacity

onmax:
## Summary

- add opt-in adaptive admission beneath the existing hard concurrency cap and FIFO queue
- add a shared Node process adapter using cgroup v2 memory limits, memory events, PSI, and conservative Node fallbacks
- expose safe live capacity status through inspection and agent info without changing static-capacity output
- export the adapter from both @vite-hub/agent/runtime/process and vite-hub/agent/runtime/process

## Behavior

Adaptive sampling can lower new admissions to zero without preempting active invocations. Pending calls remain bounded by the existing queue and resume after pressure clears. Sampling is deadline-bounded and cancellable, failures use a configured fallback, and one shared capacity object coordinates distinct Agent Definitions in a process.

## Verification

- Agent typecheck passed
- Agent build and publint passed
- 196 focused capacity, process, normalization, and CLI tests passed
- vite-hub build, publint, and package contract passed with 10 tests and no type errors
- exact final owner and umbrella tarballs passed a consumer-installed runtime and declaration smoke test
- equivalent pnpm-patched controller ran in the babysitter service for more than 60 minutes with the same PID, zero restarts, zero retained failures, and zero memory high/max/OOM events; it paused and resumed under CPU pressure and detected a new 39-job burst after reaching an empty queue


onmax:

PR #1152, opened 2026-08-27
feat(console): add read-only key-value inspection

onmax:
## What changed

- registers a read-only Console KV endpoint in both Vite and Nuxt and installs the configured default and named stores in the generated Nitro runtime
- adds a T3-style resizable list/detail UI with store selection, prefix filtering, lazy value reads, responsive navigation, and light/dark support
- bounds key responses to 200 by default (500 maximum) and value responses to 256 KiB without rewriting opaque key or prefix strings
- documents the sensitive-data, provider-operation, and production-access boundaries

The endpoint uses GET for bounded key listing and a JSON-body POST for value reads. It calls only `list`, `get`, and `has`; it has no write action and never calls `set`, `del`, or `clear`.

## Design evidence

The design follows the established KV-console pattern of choosing a namespace, filtering keys, and loading one selected value, as seen in [Redis Insight](https://redis.io/docs/latest/develop/tools/insight/). It keeps ViteHub's existing Console shell instead of introducing a separate dashboard language. Provider reads are explicit because read-only inspection can still incur operations, consistent with [Upstash's command-count guidance](https://upstash.com/docs/redis/troubleshooting/command_count_increases_unexpectedly).

## Contract coverage

- Vite and Nuxt generated plugins register the handler and preserve default plus named stores.
- KV-only Console builds do not install Agent journals or Agent read endpoints.
- Runtime tests cover default/named stores, stable filtering, missing and stored-null values, JSON/text/Unicode formatting, byte truncation, opaque identifiers, invalid methods/stores/limits, and concurrent project isolation.
- The packed package proves the distributed handler exists and is registered.
- A local distributed-package Nitro fixture verified real store reads, filtering, store switching, mobile navigation, and light/dark rendering. External managed-provider credentials and a full Nuxt browser fixture remain outside this local proof.

## Verification

- `TMPDIR=/var/tmp pnpm --filter vite-hub exec vp test -- --run test/console-kv.test.ts test/console.test.ts test/console-sections.test.ts test/nuxt.test.ts test/package.test.ts test/vite.test.ts` — 202 tests passed, no type errors
- `pnpm --filter vite-hub build` — passed, including `publint`
- `vp lint` across the changed Console, integration, build, and test files — passed


onmax:
@codex review

onmax:
@codex review

onmax:
Maintainer direction: keep read-only key enumeration and add a provider-honest paginated listing primitive in this PR at the owning `@vite-hub/kv` boundary.

Expose an obvious public contract equivalent to `list({ prefix?, limit, cursor? }) -> KVResult<{ keys, cursor? }>` with provider cursors treated as opaque. Preserve `keys()` as the explicitly exhaustive operation. Cloudflare KV, Upstash, Deno KV, and filesystem-backed stores must each stop provider work once the requested page is filled; response-only slicing is not sufficient. The Console must consume this bounded primitive, use its existing 200 default / 500 maximum, and carry the cursor for subsequent pages.

Resolve the current `main` conflict after establishing that contract, keep the endpoint read-only, and retain the documented cost and sensitive-data boundaries. This is the authorized direction needed to remove the generated blocker and continue the stack root.

PR #1151, opened 2026-08-27
feat(agent): support optional MCP servers

onmax:
## Summary

Allow an MCP server entry, or its per-invocation resolver, to be `false`, `null`, or `undefined`.

This removes the all-or-nothing setup Quiver hits today: an Agent can omit an integration whose optional credentials are unavailable without rebuilding its Capability list or disabling unrelated MCP servers. ViteHub skips only that entry and does not import or create an MCP client for it.

Configured-but-broken servers remain failures. Resolver errors, malformed configs, authentication and connection failures, tool drift, and cleanup errors are not treated as optional absence.

The docs show the matching Server Env requirement: credentials used for optional servers must themselves be declared with `optional: true`.

## Verification

- `pnpm --filter @vite-hub/agent exec vp test run test/mcp-capability.test.ts` — 18 tests passed
- `pnpm --filter @vite-hub/agent exec vp test run test/nuxt-runtime-output.test.ts` — packed Nuxt consumer passed
- `pnpm exec vp run -t @vite-hub/agent#build` — build and publint passed
- `pnpm --dir docs typecheck` — passed
- Three independent pre-PR reviews covered runtime/lifecycle, public API and consumer DX, and the affected contract matrix; all findings were addressed and re-reviewed

The runtime coverage includes static and async absent values, mixed present/absent servers, absence changing between invocations, lazy MCP imports, malformed and thrown failures, and owned-client cleanup.

PR #1154, opened 2026-08-27
feat(agent): add Executor capability

onmax:
## Summary

- add a first-class `executor()` Agent Capability for one exact Streamable HTTP MCP endpoint
- resolve connection options once per Agent Invocation so credentials can rotate without rebuilding the Agent
- share MCP discovery, integrity, metadata redaction, ownership, and cleanup behavior with `mcp()`
- document the real authentication boundary: Executor owns upstream integration credentials and policy; ViteHub receives only the gateway credential

## DX gap

Quiver currently repeats authenticated MCP URL and header setup for integrations such as Airtable and PostHog. This collapses that application plumbing to:

```ts
executor({
  apiKey: useServerEnv().executorApiKey,
  url: "https://executor.sh/acme/mcp",
})
```

Optional resolvers may return `false`, `null`, or `undefined`, so an unavailable connection contributes no tools and does not load the MCP runtime. A provided credential requires HTTPS; anonymous HTTP remains available for trusted local endpoints.

This deliberately does not add secret storage, credential rotation APIs, Codex or GitHub machine identity, Executor provisioning, or Console UI. Those remain separate primitives rather than hidden behavior in an Agent Capability.

## Runtime guarantees

- Agent Invocation abort and a configurable 30-second default initialization timeout reach the shipped MCP client
- owned clients close after success and every failure path; multi-client cleanup attempts every close
- authorization headers and URL userinfo, query, and fragment data are excluded from Capability and tool metadata
- `execute` is exposed as `executor`; auxiliary tools use the `executor_` prefix
- optional tool fingerprints block added or changed definitions before model execution
- `@ai-sdk/mcp` is pinned to 2.0.36, the first installed version used here that honors initialization timeout and abort options

## Verification

- `pnpm install --frozen-lockfile --ignore-scripts`
- focused MCP and Executor tests: 42 passed
- packed Nuxt runtime consumer: 1 passed
- `@vite-hub/agent` build and publint
- lint: no errors
- `git diff --check`
- three independent final reviews: contract matrix, consumer DX, and Executor/auth source accuracy; no actionable findings

The Agent typecheck still reports the existing `test/example.test.ts` `string | AddressInfo` error. No new source or declaration errors were reported.

Hosted behavior was verified against Executor source at `4c9e755` and the real `@ai-sdk/mcp` client, but not with a live Executor Cloud credential.

Depends on #1151.

### DIFF
--- a/docs/content/docs/capabilities/executor.md
+++ b/docs/content/docs/capabilities/executor.md
@@ -1,0 +1,87 @@
+---
+title: Executor
+description: Connect an Agent to an Executor tool catalog through one authenticated MCP endpoint.
+navigation.icon: i-lucide-plug-zap
+---
+
+Use `executor()` when an Agent should use integrations configured in [Executor](https://executor.sh/) without owning each integration credential.
+ViteHub connects to one streamable HTTP MCP endpoint. Executor keeps the Airtable, PostHog, GitHub, and other upstream credentials and policies behind that endpoint.
+
+```ts [server/agents/support.ts]
+import { useServerEnv } from '#vitehub/env/server'
+import { defineAgent } from 'vite-hub/agent'
+import { executor } from 'vite-hub/agent/capabilities'
+
+export default defineAgent({
+  driver: { model },
+  capabilities: [
+    executor({
+      apiKey: useServerEnv().executorApiKey,
+      url: 'https://executor.sh/acme/mcp',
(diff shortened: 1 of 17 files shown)


onmax:
@codex review

onmax:
@codex review

PR #1156, opened 2026-08-27
feat(console): add read-only Schedule inspection

onmax:
## Stack

Depends on #1155. This is the fourth layer in the Console inspection stack:

1. #1152 — KV
2. #1153 — Workflow
3. #1155 — Queue
4. this PR — Schedule
5. #1158 — Blob

## Summary

- add read-only Schedule navigation and list/detail Definition inspection to standalone Vite and Nuxt Consoles
- discover static cron expressions through the Schedule-owned build boundary without evaluating application modules
- show Definition kind, source, project-relative file, UTC timing, and whether a runtime target permits runtime Schedules
- state the current boundary in-product: runtime-created Schedules and run history are not part of this build-time catalog

## Contract

Schedule inspection is derived from the same Definition discovery and static cron analysis that provider output uses. It does not invent execution state, import handlers, or expose mutation controls. The Console catalog is generated on initial configuration and serialized refreshes, with the existing project-root and runtime-realm isolation preserved.

## Evidence

The list/detail shape follows the operational-console pattern used by Temporal and Trigger.dev: scan a compact schedule catalog, then inspect trustworthy timing and provenance fields. Temporal informed the strict separation between definitions and execution history; Trigger.dev informed putting cron and time-zone context beside the schedule identity.

- https://docs.temporal.io/web-ui
- https://trigger.dev/docs/management/schedules/retrieve
- https://trigger.dev/changelog/scheduled-task-timezones

## Verification

- `pnpm --filter @vite-hub/schedule exec vp test -- --run test/discovery.test.ts test/provider-output.test.ts test/package.test.ts` — 56 passed, no type errors
- `pnpm --filter vite-hub exec vp test -- --run test/console-kv.test.ts test/console-definitions.test.ts test/console.test.ts test/console-sections.test.ts test/nuxt.test.ts test/package.test.ts test/vite.test.ts` — 212 passed, no type errors
- `pnpm --filter @vite-hub/schedule build` — package build and publint passed
- `pnpm --filter vite-hub build` — Console build, package build, and publint passed
- `pnpm exec vite-doctor scan . --since t3code/console-queue-inspection` — 100/100
- packaged Cloudflare consumer exercised at `/_vitehub/schedules` in light and dark modes; catalog/API values matched and the browser reported no page errors

PR #1155, opened 2026-08-27
feat(console): add read-only Queue inspection

onmax:
## Summary

- add a read-only Queue section to the shared Console definition catalog
- discover Queue Definitions at build time without evaluating their modules
- show project-relative Definition identity and source metadata in the existing responsive list/detail shell
- omit backlog, message, delivery, retry, and dead-letter controls until ViteHub owns a provider-independent inspection contract

## Product evidence

Cloudflare and Vercel both make backlog amount, message age, throughput, and consumer state central to Queue operations. Those products also expose provider-specific mutation controls. This layer adopts their drill-down navigation but does not synthesize operational state from Definition configuration or surface write controls.

- https://developers.cloudflare.com/queues/observability/metrics/
- https://developers.cloudflare.com/queues/examples/list-messages-from-dash/
- https://vercel.com/docs/queues/observability

## Verification

- `TMPDIR=/var/tmp corepack pnpm --filter @vite-hub/queue exec vp test -- --run test/discovery.test.ts test/package.test.ts`
- `TMPDIR=/var/tmp corepack pnpm --filter vite-hub exec vp test -- --run test/console-kv.test.ts test/console-definitions.test.ts test/console.test.ts test/console-sections.test.ts test/nuxt.test.ts test/package.test.ts test/vite.test.ts`
- `TMPDIR=/var/tmp corepack pnpm --filter @vite-hub/queue build`
- `TMPDIR=/var/tmp corepack pnpm --filter vite-hub build`
- `VITE_DOCTOR_SINCE=t3code/console-workflow-inspection vp run doctor:typescript` — 100/100
- distributed standalone Cloudflare fixture checked in the browser with a discovered Queue Definition

Stacked on #1153.

PR #1153, opened 2026-08-27
feat(console): add read-only Workflow inspection

onmax:
## Summary

- add a read-only Workflow section to the configuration-aware Console
- discover Workflow Definitions at build time without evaluating their modules
- reuse one responsive list/detail shell for definition-backed Console primitives
- show project-relative source metadata, Agent identity, and ordered steps when known
- keep run history out of the UI until ViteHub owns a provider-independent listing contract

## Product evidence

Temporal, Inngest, AWS Step Functions, and Vercel all support a stable list-to-detail inspection pattern. Their run timelines depend on execution-history contracts ViteHub does not currently share across providers, so this layer adopts the navigation pattern while limiting the detail to trustworthy Definition metadata.

- https://docs.temporal.io/web-ui
- https://www.inngest.com/docs/platform/monitor/traces
- https://docs.aws.amazon.com/step-functions/latest/dg/concepts-view-execution-details.html
- https://vercel.com/workflows

## Verification

- `TMPDIR=/var/tmp corepack pnpm --filter @vite-hub/workflow exec vp test -- --run test/definition.test.ts test/package.test.ts`
- `TMPDIR=/var/tmp corepack pnpm --filter vite-hub exec vp test -- --run test/console-kv.test.ts test/console-definitions.test.ts test/console.test.ts test/console-sections.test.ts test/nuxt.test.ts test/package.test.ts test/vite.test.ts`
- `TMPDIR=/var/tmp corepack pnpm --filter @vite-hub/workflow build`
- `TMPDIR=/var/tmp corepack pnpm --filter vite-hub build`
- distributed standalone Nitro fixture checked in light desktop and dark mobile layouts, including the responsive sidebar

Stacked on #1152.

PR #1158, opened 2026-08-27
feat(console): inspect Blob metadata

onmax:
## Stack

Depends on #1156. This is the fifth layer in the Console inspection stack:

1. #1152 — KV
2. #1153 — Workflow
3. #1155 — Queue
4. #1156 — Schedule
5. this PR — Blob
6. #1159 — Database

## What changed

Add a read-only Blob section to the Console. It lists configured stores with bounded, cursor-aware pagination and prefix filtering, then renders object metadata in the same dense list/detail shell as the existing Console primitives.

The inspection boundary calls only `BlobStorage.list`. It never downloads object contents, invokes Blob mutations, or returns provider URLs. The docs call out that listing can still incur provider requests and cost.

This also pins the standalone Console to its own `vue-router` instance and passes the router adapter expected by Nuxt UI. Visual verification exposed that the packaged app could otherwise resolve two router instances and crash every Console route.

## Verification

- 133 focused Console, Nuxt, package, and Blob tests pass with no type errors
- Console build and package validation pass, including publint
- Changed files pass lint and `git diff --check`
- Vite Doctor: 100/100, no findings
- Browser QA covered populated, filtered, empty, desktop, mobile, light, dark, and mobile-sidebar states
- API regression coverage proves URL redaction, store isolation, pagination bounds, validation, and the absence of `get`, `head`, `serve`, `sign`, `put`, and `del` calls


onmax:
@codex review

onmax:
@codex review

onmax:
Superseded by #1205, which consolidates the serial Console inspection stack onto current main and preserves this branch for history.

PR #1197, opened 2026-08-30
fix(agent): make trigger history explicit

onmax:
## Summary

- Send only the current chat message by default.
- Keep adapter `threadHistory` backfill/cache separate from model-visible `triggerHistory`.
- Let each Agent or Channel opt into an explicit count-bounded thread window, with an optional age bound.
- Document the contract and update history-dependent tests to opt in.

## Behavior

An enabled window is now explicit and bounded:

```ts
triggerHistory: {
  source: 'thread',
  maxMessages: 20,
  maxAgeMs: 30 * 60 * 1000,
}
```

`maxMessages` includes the current message. `maxAgeMs` removes a stale prefix relative to that message while preserving the current message as the final input. Missing timestamps start a new history boundary.

This is intentionally a type-level breaking change for enabled windows: `maxMessages` is required. Omitted or invalid runtime history settings fail closed to the current message.

## Why

History retrieval and model context are different decisions. A Channel may need thread backfill for transcripts or delivery state without automatically sending that entire cache to the model. Keeping selection explicit gives one-thread Channels a current-only default while allowing conversational products to choose their own count and recency policy.

## Verification

- `pnpm vp run -t @vite-hub/agent#build`
- `pnpm vp test run test/chat-message-input.test.ts test/channel-instructions.test.ts` (38 passed)
- Agent Dev Loop explicit-history integration (1 passed)
- adapter webhook current-only regression (1 passed)
- `git diff --check`

The full local Agent suite completed with 2,549 passing and 2 skipped tests. Nine unrelated tests failed in the local harness: six generated-route fixtures hit an H3 mock-request error, one published-package fixture inherited `/tmp/package.json`'s npm package-manager setting, and two provider coordination tests were timing-sensitive under the full-suite load.

### DIFF
--- a/docs/content/docs/agents/chat-history-sessions.md
+++ b/docs/content/docs/agents/chat-history-sessions.md
@@ -8,6 +8,8 @@ icon: i-lucide-messages-square
 
 Chat History is the ordered set of prior messages eligible for one chat invocation. A Chat Session selects the host-visible conversation boundary. Neither is durable Agent Memory.
 
+ViteHub sends only the current message unless `triggerHistory` explicitly enables a bounded window. Adapter `threadHistory` controls backfill and caching but does not select model input.
+
 | Need | Use |
 | --- | --- |
 | Continue the visible thread | Thread-backed Chat History |
@@ -32,6 +34,7 @@ export default defineAgent({
       concurrency: 'queue',
       lockScope: 'thread',
       triggerHistory: {
+        maxAgeMs: 30 * 60 * 1000,
         maxMessages: 20,
         source: 'thread',
       },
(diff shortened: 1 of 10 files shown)


onmax:
@pullfrog Please review this PR at exact head `c9fec9edbd70a976ce5ac6d13d25a38da2a4c63c` and submit a GitHub review.

onmax:
New direction: keep this PR at the model-input selection boundary.

The intended contract is still:

- omitted `triggerHistory` sends only the current message;
- `threadHistory` may backfill or cache messages, but does not implicitly select model input;
- an explicit thread window requires `maxMessages`, with optional `maxAgeMs`;
- the current message remains the final selected message.

Please return the implementation close to the original `71598fe` change and preserve the existing webhook lifecycle around it. In particular, this PR should not change session identity, invoker resolution, authorization order, attachment hydration, durable-history merging, id-less collision policy, or provider cache behavior. Select model-visible messages from the ordered input only after those existing owners have done their work.

The current +637/-102 diff and 38-commit repair chain show that the change crossed those ownership boundaries. Revert the follow-up machinery rather than continuing to repair it hunk by hunk. Keep focused contract tests for current-only default, explicit count and age windows, and independence from `threadHistory`. If the investigation found real defects in sessions, attachments, authorization, or durable history, move each one to a separately named PR with its own lifecycle proof.

Target shape: the public type, `chat-message-input.ts`, docs, and focused tests. Preserve adjacent runtime behavior unchanged.

PR #1200, opened 2026-08-30
fix(workflow): stabilize staged provider imports

onmax:
Generates Workflow registry imports as build-time file URLs and provider runtime imports as stable package specifiers. When Provider Output stages and later relocates generated files, discovered handlers and steps remain pinned to the retained build snapshot instead of the deleted output tree or pnpm store paths.

Configured server directories outside the retained source closure keep their original paths. Normal non-staged discovery keeps the same source root.

This covers both Cloudflare and Vercel generated entries.

Test plan:
- pnpm --filter @vite-hub/workflow exec vp test test/registry.test.ts
- pnpm --filter @vite-hub/workflow exec vp test test/vite-output.test.ts -t "publishes staged generated Workflow entries|preserves staged Workflow imports from configured external server directories"
- pnpm --filter @vite-hub/workflow typecheck

Downstream proof: vite-hub/strata@9ddb58d completes a Cloudflare build and live Workflow execution with the equivalent pnpm patch.


onmax:
@pullfrog Please review this PR at exact head `c52df54111215af7873d96f38ecf79a2d344ae45` and submit a GitHub review.

PR #1202, opened 2026-08-30
feat(kv): add atomic operations

onmax:
ViteHub KV users on Upstash can now consume one-time values and increment fixed-window counters atomically. Each method runs as one provider operation, so concurrent requests cannot change the value between a read and write.

This also lets ViteHub integrations implement the [`getAndDelete()` and `increment()` methods in Better Auth 1.7 secondary storage](https://better-auth.com/docs/concepts/database#secondary-storage) without adding a separate storage backend. Nuxt Better Auth users reported the same provider gap in [nuxt-modules/better-auth#423](https://github.com/nuxt-modules/better-auth/issues/423), and [nuxt-hub/core#927](https://github.com/nuxt-hub/core/pull/927) adds the matching NuxtHub bridge.

## Driver behavior

| Driver | Behavior | Provider reference |
| --- | --- | --- |
| Upstash | Both methods are available | [Upstash `GETDEL`](https://upstash.com/docs/redis/sdks/ts/commands/string/getdel) · [Upstash `EVAL`](https://upstash.com/docs/redis/sdks/ts/commands/scripts/eval) |
| Deno KV | Both methods are unavailable | [Deno KV transactions](https://docs.deno.com/deploy/kv/transactions/) · [Deno KV key expiration](https://docs.deno.com/deploy/kv/key_expiration/) |
| Cloudflare KV | Both methods are unavailable | [How Workers KV works](https://developers.cloudflare.com/kv/concepts/how-kv-works/) |
| `fs-lite` | Both methods are unavailable | No provider operation is available |

Upstash uses `GETDEL` for one-time reads. Its counter runs `EXISTS`, `INCR`, and the initial `EXPIRE` in one Lua script. Checking existence before incrementing matters because an existing counter may contain zero and must keep its current expiry. ViteHub also validates the TTL, normalizes keys like ordinary KV methods, and decodes atomic reads through the same path as `get()`.

Deno KV can transact the value update, but each write accepts a new relative expiry. Preserving the original fixed window would require a hidden deadline key and cleanup on every write or deletion. This PR leaves the existing Deno KV paths unchanged. Cloudflare KV and local files also reject the methods because they cannot provide the same atomic contract.


onmax:
@pullfrog Please review this PR at exact head `58321de4edd56e9eabd58dc75e1bc4e8e4f7b8be`. Focus on the changes since the last reviewed head and report any current-head findings.

PR #1217, opened 2026-09-01
fix(docs): trim ViteHub logo canvas

onmax:
Reduces the horizontal SVG canvas from 1296 to 900 units so the ViteHub logo no longer includes a large empty area on its right. The mark, wordmark placement, height, and accessible label remain unchanged.

## Test plan

- `git diff --check origin/main...HEAD`
- Confirmed the explicit width and viewBox use the same 900 × 304 canvas


onmax:
merge it pls

PR #1227, opened 2026-09-01
fix(console): refresh standalone assets after deploy

onmax:
The standalone Console shell used stable asset URLs, so a tab or browser cache could keep the pre-deploy two-column trace bundle after the server had started serving the new bottom splitter.

Emit content-hashed filenames for the Console CSS and JavaScript, then insert those exact filenames into the packaged standalone shell. A rolling deployment cannot serve old bytes for a new asset URL. The existing Nuxt UI vertical splitter remains the owning layout: waterfall above, selected-span details below, draggable horizontal boundary, and scrollable detail content.

Verified locally:
- targeted lint
- TypeScript doctor
- diff check

The focused test and package typecheck require workspace build output, so exact-head CI provides those checks.

### DIFF
--- a/packages/vite-hub/console.vite.config.ts
+++ b/packages/vite-hub/console.vite.config.ts
@@ -42,10 +42,10 @@ export default defineConfig({
       input: resolve(clientRoot, "main.js"),
       output: {
         assetFileNames: asset => asset.names.some(name => name.endsWith(".css"))
-          ? "console.css"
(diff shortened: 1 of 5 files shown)


onmax:
Addressed the rolling-deployment concern by content-addressing the standalone Console CSS and JavaScript filenames. The package test now verifies that the built shell references the exact hashed assets and contains no unresolved placeholders.

PR #1252, opened 2026-09-03
ci: remove stale Fallow job

onmax:
## Summary

Remove the main-only Fallow CI job after Fallow was removed in #1251.

## Verification

- `git diff --check`
- confirmed no remaining Fallow references in CI or root task configuration


onmax:
also remove the .fallow folder

PR #1347, opened 2026-09-08
fix(workspace): persist local file metadata

onmax:
The local Workspace store kept file media types and metadata only in process memory, so ownership and provenance disappeared after restart. This persists each file's metadata atomically in a bounded, hidden sidecar tree under the reserved `.vitehub/file-metadata` directory inside the Workspace root and reloads it for read, stat, list, snapshots, byte writes, streamed writes, and removal.

Sidecars use `<workspace-path>/metadata.json`, which keeps valid paths such as `a` and `a.json/b` independent. Tests cover restart recovery, invisible sidecars, recursive cleanup, suffix-related paths, and unchanged/streamed content.

Validation: 94 local-store tests passed with a temporary source alias for the unbuilt runtime dependency; focused oxlint passed. Package typecheck was attempted but cannot complete without dependency output declarations. No local builds were run.


onmax:
**Verdict: revise**

Head `1f004e8e843afdb9e167606412e9f42d3ae717b1` implements the stated durability outcome, but the diff has expanded to 30 files and 3,059 additions across Source resolution and non-local stores. The title and body scope the intent to persisting local Workspace file metadata. Reconcile with the current `main`, retain only contract changes required for that outcome, and verify the reduced exact-head diff before merge.

Evidence: PR title/body, `git diff --stat origin/main...HEAD`, and current base/head ancestry checked 2026-09-10.

onmax:
@codex review

onmax:
Release audit follow-up: the local metadata sidecar still has a cross-instance lost-update race. Each LocalWorkspaceStore loads metadata once and rewrites its stale in-memory map (`packages/workspace/src/storage/local.ts:427-430,451-462`). Two workers sharing a root can overwrite each other's metadata; this affects source/materialization provenance. Please add an inter-process lock or atomic read-modify-write regression test before treating this PR as complete.

onmax:
Closing because the prepared source checkout is unavailable: the commit objects exist, but the worktree/index contains zero tracked files, preventing verification or repair. Please reopen from a restored checkout.

PR #1364, opened 2026-09-09
fix(agent): defer provider readiness until input commands run

onmax:
## Problem

Input commands that return a response should be able to handle a request before provider readiness checks run. A provider can be unavailable because of quota while a command such as `/debug` only needs to inspect an existing invocation and reply. Starting the readiness check in parallel means that unavailable provider state can fail the command path even though the command does not need the provider.

## Change

Defer provider readiness evaluation until capability input preparation completes. If input handling returns a response, readiness is never evaluated. If input handling continues, readiness is checked before agent input hooks and driver execution as before.

This preserves the existing `context.reply()` response short circuit and makes the ordering explicit at the invocation boundary.

## Validation

The Vitest binary is not present in this checkout, so the focused runtime tests could not be run locally.


onmax:
missing docs and tests?

PR #1375, opened 2026-09-10
refactor(blob): share provider dependency mappings

onmax:
Blob maintains the same provider SDK mapping separately for install guidance and deployment packaging, plus a third list for package externals. Use the existing filesSdkDriverPeers map for all three so adding or changing a provider's requirements has one owner.

Removes 33 net lines. Public exports, emitted declarations, SDK ordering, optional-peer diagnostics and native Cloudflare R2 behavior are unchanged.

Validation:

- Full Blob package suite: 247 tests passed, including type tests and packed private Vercel Blob / Cloudflare R2 consumer builds without provider SDKs at the consumer root.
- Package build, publint and typecheck passed.
- Emitted declarations match the baseline byte for byte.
- Repository lint, TypeScript doctor and git diff --check passed.

Independent PR against main; it does not depend on the dependency cleanup in #1374.

PR #1374, opened 2026-09-10
chore: trim Sandbox and Source runtime dependencies

onmax:
Sandbox installs packages it does not use at runtime, and Source declares Effect as a runtime dependency even though it bundles it. Remove Sandbox's h3, ufo and scule declarations, preserve diagnostic capitalization locally, and move Sandbox's private Rollup type dependency and Source's bundled Effect to devDependencies. Remove the unused catalogs and update the lockfile.

Public exports and emitted declarations are unchanged. This reduces direct runtime requirements; other packages can still install these dependencies transitively.

Validation:

- Full Sandbox suite: 344 passed. Full Source suite: 127 passed.
- Both package typechecks and builds, including publint, passed.
- Fresh emitted declarations match the baseline byte for byte. Built JavaScript and declarations have no imports from the removed dependencies.
- Full workspace build, repository lint, TypeScript doctor and Knip catalog check passed.
- Frozen lockfile installation passed.
- Root contracts: 349 passed on the full run; the remaining license-packaging test passed when rerun alone after a 30-second timeout. The updated Effect ownership contract accepts the pinned catalog in any dependency group and checks that Source emits no external Effect imports.
- The broader packed-consumer suite is running locally; its result will be added when complete.

PR #1376, opened 2026-09-10
refactor(shell): share command parsing and inspection metadata

onmax:
Shell execution, command analysis and Workspace inspection duplicated quote/escape handling and command splitting. Share those helpers inside the Shell package and derive command names and pipeline flags from the same segments. This removes 105 net implementation lines.

Preserve execution's malformed-quote errors and trailing backslashes, inspection's tolerance of incomplete input, separator/pipeline metadata, and the existing sh-syntax parser and diagnostics. Public exports and emitted declarations are unchanged.

Validation:

- Full Shell suite: 29 tests passed, including seven new provider-forwarding and analyzer regressions.
- Before/after comparisons passed for 20,011 deterministic tokenizer, segment and Workspace inspection inputs, plus 20 public analyzer cases. Comparisons included errors, malformed quotes, escapes, separators, pipelines and incomplete commands.
- Package build, publint and typecheck passed. Emitted public declarations match the baseline byte for byte.
- Repository lint, TypeScript doctor and git diff --check passed.
- Cloudflare forwarding was checked with the client interface; no live provider run was needed for this shared parsing change.

Independent PR against main; it does not depend on #1374 or #1375.

### DIFF
--- a/packages/shell/src/command/analyze.ts
+++ b/packages/shell/src/command/analyze.ts
@@ -1,3 +1,5 @@
+import { splitShellCommandSegments } from "./parse.ts"
+
 import type { ShellAnalyzeOptions, ShellAnalyzeResult } from "../runtime/types.ts"
 import { shellErrorDiagnostics } from "../error-diagnostics.ts"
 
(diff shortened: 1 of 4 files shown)


onmax:
> Direction: proceed
>
> Evidence: PR title/body match shared Shell parsing refactor; changed files are limited to Shell parser/analyzer/workspace inspection plus regressions. Existing review reports no new issues.

onmax:
@codex review

onmax:
Release audit follow-up: concurrent Shell `startProcess()` calls can bypass `maxProcesses` while provider startup is pending. I opened #1379 with the fix and regression test. Please avoid duplicating that change here unless this branch is the intended owner.

PR #1377, opened 2026-09-10
fix(ui): prevent raw HTML execution in agent Markdown

onmax:
## Problem
Agent message Markdown was rendered with raw HTML enabled. Untrusted assistant or user text could produce executable script/event-handler markup through `AgentMarkdown` and `AgentMessageParts`.

## Change
Disable raw HTML parsing in the shared AgentMarkdown renderer and add an SSR regression test covering script and event-handler input.

## Validation
- `pnpm --dir packages/ui exec vp test test/ssr.test.ts`


onmax:
@codex review

onmax:
Closing because the prepared source checkout is unavailable in this worktree, so the requested verification and repair cannot be performed.

PR #1386, opened 2026-09-13
feat(agent): support preset definitions

onmax:
## What changed

Add `preset` as the public name for agent composition while reusing the existing `extends` layer resolver internally. Presets inherit the parent definition, keep existing merge semantics, and reject ambiguous `extends` plus `preset` configuration.

## Why

This gives applications and published agent packages a stable composition entry point without adding a second merge system or making the core agent package opinionated.

## Validation

- `corepack pnpm --dir packages/agent exec vitest run test/agent-layers.test.ts`


onmax:
Superseded by the stronger named preset registry in #1387.

PR #1421, opened 2026-09-19
refactor(agent)!: replace askweb with @agntn/web

oritwoen:
askweb is a dead end: 0.2.0 is its last release, the code lives on as `@agntn/web` in the agntn org (my package under both names). `webSearch()` tool mode loads `@agntn/web` instead, so apps on tool mode have to swap the install. All seven documented provider names are in the new builtin set, plus five more.

Model: Claude Opus 5
Harness: Claude Code


onmax:
Thank you!

onmax:
Thank you!

